PR #8, opened 2023-08-25
add attachments

jchris:
this allows saving and loading of File objects


jchris:
how much work would it be to sidecar them so they don't slow down initial app load times? since the file() method is async, we can get the car file for the attachment then.

this would mean we need to keep the attachment car in the attachment doc, and we can't compact attachments easily, but maybe worth it. We can put all files from same bulk update into one car.

Which means there's a good reason to expose bulk at the db layer, b/c a multi file upload can become a single car

jchris:
LFG

Screen Shot 2023-08-24 at 11 02 03 PM

PR #10, opened 2023-08-29
w3clock

jchris:
Currently works for uploading and downloading car files to web3.storage, the w3clock (for tracking latest car file) integration requires some finalizing:

- [x] fix the fetch.node thing in ipfs-utils (hardedcoded `const fetch = require('./fetch.node')` in. `../../node_modules/.pnpm/ipfs-utils@9.0.14`
- [ ] apply https://github.com/alanshaw/pail/pull/20
- [ ] get ability to advance clock, maybe via: https://github.com/web3-storage/w3up/issues/876
- [x] revalidate build due to commonjs importer

### DIFF
--- a/packages/fireproof/src/connect-web3.ts
+++ b/packages/fireproof/src/connect-web3.ts
@@ -2,9 +2,11 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 import { create, Client } from '@web3-storage/w3up-client'
+import * as w3clock from '@web3-storage/clock/client'
 
 import { Connection, DownloadFnParams, UploadFnParams } from './types'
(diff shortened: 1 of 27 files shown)


packages/fireproof/src/connect-web3.ts
```diff
@@ -0,0 +1,132 @@
...
+      // @ts-ignore
+      const ag = this.client._agent
+      const issuer = ag.issuer
+      console.log('issuer', issuer, issuer.signatureAlgorithm, issuer.did())
+
+      if (!issuer.signatureAlgorithm) { throw new Error('issuer not valid') }
+
+      const claims = await this.client.capability.access.claim()
+
+      const delegated = await clock.delegate({
```
jchris, line 71: this is where we attempt to prove we have ability to advance the clock eg `{ can: 'clock/*' }`

PR #18, opened 2023-09-15
changes to file testing

dhruvsoni1802:


jchris:
it would be good if you can base you work on https://github.com/fireproof-storage/fireproof/pull/16/files which does not have the changes to settings.js -- also note I changed your formatting to match the rest of the file

PR #36, opened 2023-10-01
fix #21

### LINKED ISSUE #21: feature: logical clock history and logical snapshots
The `changes` api should return the resulting clock head of each change in the history. User agents can use this to load the database at any snapshot.

Keep in mind:

- [ ] loading a snapshot should disconnect any remotes, we don't want time travel to be contagious
- [ ] the desired behavior can probably be accomplished today by mutating the `db._crdt.clock.head` but we want a safer API
- [ ] this is distinct from an API for loading the database from storage snapshots, this is more about navigating the logical clock history, that is more about storage snapshot and rollback. they are similar because currently each storage snapshot corresponds to each database operation, but that might change as we add more indexing options, etc

dhruvsoni1802:


jchris:
this looks basically great. reviewing it now

jchris (approved):
This is awesome, great clean patch.

PR #41, opened 2023-10-29
Fireproof and React Native

boorad:


jchris:
This looks ready for review -- any gotchas someone should know about before trying to set up the dev environment locally? Is it feasible to add the CI/CD to the github actions (maybe its already done)?

boorad:
will try again in #107

PR #51, opened 2023-12-24
PartyKit storage

jchris:
this might fix the decode problems

### DIFF
--- a/packages/connect-partykit/src/server.ts
+++ b/packages/connect-partykit/src/server.ts
@@ -6,6 +6,14 @@ type PartyMessage = {
   parents: string[]
 }
 
+const CORS = {
+  'Access-Control-Allow-Origin': '*',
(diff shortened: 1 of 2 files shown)


packages/connect-partykit/src/server.ts
```diff
@@ -19,48 +27,31 @@ export default class Server implements Party.Server {
...
-      }
-       else if (request.method === 'GET') {
-        const carArrayBuffer = (await this.party.storage.get(
-          `car-${carId}`
-        )) as Uint8Array;
+      } else if (request.method === 'GET') {
+        const carArrayBuffer = (await this.party.storage.get(`car-${carId}`)) as Uint8Array
         if (carArrayBuffer) {
-          return json(carArrayBuffer);
+          return new Response(carArrayBuffer, { status: 200, headers: CORS })
```
jchris, line 48: if the patch doesn't work, try reverting to `main` and just changing this one line @valorant-dhruv

jchris:
the work has moved to another branch

PR #58, opened 2024-01-07
build: updates root-level project scripts to support recursive actions

thedanchez:
This PR is a quality of maintainer/contributor life improvement to add top-level scripts that run actions such as:
- `build`: recursively run all `build` scripts defined within packages
- `build:clean`: clears all `dist` directories from packages
- `clean`: clears all `node_modules` directories from packages
- `clean:all`: clears all `dist` and `node_modules` from workspace
- `test`: recursively run all `test` scripts defined within workspace packages

And package-scoped versions of the above actions for:
- `@fireproof/encrypted-blockstore`
- `@fireproof/core`
- `use-fireproof`

In doing this effort, many things came up thanks to TypeScript catching problems after renaming some `types.d.ts` files to `types.ts` among other things. 

This PR also adds a `.nvmrc` file to define the expected Node version for the project (managed by [NVM](https://github.com/nvm-sh/nvm)). As a maintainer/contributor, assuming `nvm` is installed, all you have to do is execute `nvm use` and it will set everything up. It is currently set to the latest Node 2010.0 LTS. (NOTE: Node 16 is end of life so we should not be using that anywhere). With Node 18+, the `fetch` API is natively built into the runtime, so we should be able to move away from things like `cross-fetch` but I felt we could leave that shift to another PR to not further pollute this one.

At the very least, having these scripts I think should provide more confidence going forward now that we will be able to see all packages being built and tested successfully alongside potentially adding the success runs of these scripts to GitHub Action workflows.

## Remaining TODOs:
- Fix build issues with `examples/todomvc`
- Fix test script issues with `use-fireproof` React package


thedanchez:
There seems to be a lot of issues with the `todomvc` example -- my gut is telling me from the looks of it, that example seems pretty out of date with the current state of the `@fireproof/core` APIs among other things. Is that a fair characterization of the matter or is there something else I'm missing? cc @jchris

jchris:
yes the examples need rehab -- there are better examples in other repos. It'd be worth a distinct cleanup effort, probably more along the lines of replacing them or making that folder into links to up to date repos.

packages/connect-ipfs/src/connect-ipfs-account.ts
```diff
@@ -58,8 +59,8 @@ export class AccountConnectIPFS extends DatabaseConnectIPFS {
   constructor() {
     super()
     this.accountDb = fireproof('_connect-web3')
-    const { _crdt: { blocks: { loader: accountDbLoader } } } = this.accountDb
-    this.connect(accountDbLoader!)
```
jchris, line 62: thank you!

thedanchez: No problem!

jchris:
I'm merging, and then going to do a commit to switch testing to just the core projects (as it is today) -- once we have `test:recursive` passing we can switch back. This will let us tackle that stuff as a distinct change. Thank you!

PR #61, opened 2024-01-08
build: use tsup to bundle @fireproof/encrypted-blockstore

thedanchez:
## Summary

This PR is exploring the use of `tsup` (via `tsup.config.ts`) to bundle `@fireproof/encrypted-blockstore`. 

The primary motivation behind this is to move towards a more declarative form of bundling as opposed to the current manual invocations of `tsc` and supplemental scripts contained within the `scripts` directory.

Having this in place should smooth out the experience of maintaining the key resources dedicated to bundling as a primary project task.

## Overview

The `tsup` build output generates two sub-directories:
- `dist/lib` containing all the source files in a flat structure (since we already suffix files with `-web` or `-node`)
- `dist/test` containing files to be used for testing

We can revert back to the separate `dist/web`, `dist/node` and `dist/types` directories if that desire is still there but just wanted to raise awareness to the above change regarding build output.


jchris:
The thing I can't tell from here, is does this change what gets bundled for web? The current build makes a browser bundle using webcrypto, and polyfills that on node, but maybe you have it using node-crypto here?

I wouldn't care about that as much as long as we are able to maintain our compatiblity matrix and avoid bundle size bloat. Periodically I do a size check, I'm sure we are due for a cleanup.

thedanchez:
> The thing I can't tell from here, is does this change what gets bundled for web? The current build makes a browser bundle using webcrypto, and polyfills that on node, but maybe you have it using node-crypto here?

@jchris Yeah, I was trying to make sure I kept that consistent. That's where this comes in:

```tsx
const WEB_PLUGINS = [
  polyfillNode({
    polyfills: { 
      crypto: true, 
      fs: true, 
      process: 'empty' 
    },
  })
]
```
I may be consuming it in the wrong places (or have it misconfigured) but this is my first time dealing with a polyfill in a bundle scenario so any guidance and/or direction to make sure I'm polyfilling for the right targets will be welcomed with open arms and very much appreciated 😄

jchris:
I am cool with polyfill, but the browser should get the lightest build possible, that's why today we don't ship code with fs calls to the browser, and we use web crypto. 

We could polyfill indexDB for node, but the filesystem api make sense and maybe in the long run we move to the web FS api. So the approach in the `eb-web.ts`, `eb-node.ts`, `eb-*.ts` files is necessary I think. Each of those can be swapped in for a build. And that way we can have a build with deno crypto and cloudflare kv storage. etc. Making a remix is as easy as adding a new `eb-*.ts` file.

packages/fireproof/tsconfig.json
```diff
@@ -15,6 +15,9 @@
     "declaration": true,
     "outDir": "dist",
     "rootDir": "src",
-    "forceConsistentCasingInFileNames": true
+    "forceConsistentCasingInFileNames": true,
+    "paths": {
+      "@fireproof/encrypted-blockstore/*": ["./node_modules/@fireproof/encrypted-blockstore/dist/*"]
+    }
```
thedanchez: Added `paths` here so we can get rid of those `@ts-ignore` comments in the `eb-*.ts` files when importing from the multiple entry points.

thedanchez: Decided to split out this little tidbit into https://github.com/fireproof-storage/fireproof/pull/66

thedanchez: Nvm -- brought this back in

thedanchez:
@jchris so the thing I'm trying to figure out now is why the outputs are different between what the scripts output compared to `tsup`. Looking at the bundle output of `store-node` for example gives the following.

### Using scripts

Screenshot 2024-01-15 at 1 38 24 AM

### Using `tsup`

Screenshot 2024-01-15 at 1 39 28 AM


Massive difference in LOC count: 7830 vs 320 🤯  -- but the current test suite is passing with `tsup` build artifacts. However I am concerned that the `tsup` artifacts may be missing things that they are supposed to have.

jchris:
Thank you. Merging. I can't wait to see this elsewhere in the build.

PR #53, opened 2023-12-28
feat: adds solid-js support

thedanchez:
## Summary

Say hello to `@fireproof/solid-js` -- the official SolidJS adapter for Fireproof. 

Light up your data with an embedded live database for your SolidJS web app!

## Overview

This PR adds support for SolidJS as a first-class citizen like React. The intent is to create another npm package `@fireproof/solid-js` that consumers can import like so:

```tsx
import { createFireproof } from "@fireproof/solid-js"`
```

The SolidJS hook interface (prefixed with `create` per Solid conventions) mirror its React counterpart (prefixed with `use`) with some changes to adapt to Solid -- the main one being, you receive _accessor_  functions instead of discrete values from the SolidJS hook as there are no re-renders in Solid as there are in React.

```tsx
import { createFireproof } from "@fireproof/solid-js"`

type Todo = { text: string; date: number; completed: boolean };

// You can have a global database that any Solid component can import as an alternative to Context APIs
export const TodoDB = createFireproof('TodoDB');

function App() {
  // completedTodos is an accessor/getter function (i.e. completedTodos()) instead of a value
  const completedTodos = TodoDB.createLiveQuery<Todo>('completed', { limit: 10 })
  const [todo, setTodo, saveTodo] = TodoDB.createDocument<Todo>(() => ({
    text: '',
    date: Date.now(),
    completed: false,
  }));
}
```

Other thing to note is also making heavy use of JSDoc comments to provide as much guidance to consumers of the hooks and their usage to lift the DX a bit.

PR #64, opened 2024-01-10
feat: adds further support for generics across type definitions

thedanchez:
## Summary

After merging https://github.com/fireproof-storage/fireproof/pull/63, realized there was more to do on the type enhancement front to support generics across the `@fireproof` type definitions ecosystem.

## Note

I used the SolidJS example from https://github.com/fireproof-storage/fireproof/pull/53 to test out the correctness of the type changes. I'm thinking we should probably expand the example further to flex more of the type definition areas to gain further confidence if we are wary of these changes in some regard.

### DIFF
--- a/packages/fireproof/src/database.ts
+++ b/packages/fireproof/src/database.ts
@@ -11,7 +11,8 @@ import type {
   ConfigOpts,
   MapFn,
   QueryOpts,
-  ChangesOptions
+  ChangesOptions,
+  DocRecord
 } from './types'
 import { DbResponse, ChangesResponse } from './types'
 import { EncryptedBlockstore } from '@fireproof/encrypted-blockstore'
@@ -45,7 +46,7 @@ export class Database {
     })
   }
 
-  async get(id: string): Promise<Doc> {
+  async get<T extends DocRecord = {}>(id: string): Promise<Doc<T>> {
     const got = await this._crdt.get(id).catch(e => {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
       e.message = `Not found: ${id} - ` + e.message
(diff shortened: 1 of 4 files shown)


packages/fireproof/src/types.ts
```diff
@@ -44,20 +44,19 @@ export type DocFileMeta = {
 
 export type DocFiles = Record<string, DocFileMeta | File>;
 
-type DocBody<T extends Record<string, unknown> = {}> = {
-  _files?: DocFiles;
-  _publicFiles?: DocFiles;
-} & Record<string, DocFragment> & T
+export type DocBody<T extends Record<string, unknown> = {}> = {
+  _id?: string;
+} & { [K in Exclude<keyof T, keyof DocBase>]: DocFragment } & T
```
thedanchez: I have to say, getting the `Record<string, DocFragment>` to work was rough. Had to do some `Exclude` voodoo as you see here in order for things to work properly. Thanks to the SolidJS example, I could see the affect this `DocBody` type was having upstream at the call sites.

thedanchez:
I'm wondering if I should refactor things on the generics side to extend `Record<string, DocFragment>` instead of `Record<string, unknown>`.

thedanchez:
> I'm wondering if I should refactor things on the generics side to extend `Record<string, DocFragment>` instead of `Record<string, unknown>`.

@jchris any thoughts on this?

packages/fireproof/src/database.ts
```diff
@@ -45,18 +45,18 @@ export class Database {
     })
   }
 
-  async get(id: string): Promise<Doc> {
+  async get<T extends Record<string, unknown> = {}>(id: string): Promise<Doc<T>> {
```
jchris: this `T extends Record<string, unknown>` is repeating all over. Does it make sense to name it and reuse that. Maybe DocRecord?

jchris:
I think the idea of `Record<string, DocFragment>` is good -- DocFragment might get refined over time. It would be nice to get the` _files` behavior into the typedef

packages/fireproof/src/database.ts
thedanchez: Yeah, absolutely makes sense.

packages/fireproof/src/index.ts
```diff
@@ -260,8 +263,6 @@ export class Index {
   }
 }
 
-function makeMapFnFromName(name: string): MapFn {
-  return doc => {
-    if (doc[name]) return doc[name]
-  }
+function makeMapFnFromName<T extends DocRecord = {}>(name: keyof Doc<T>): MapFn {
+  return doc => doc[name] ?? undefined
```
jchris, line 267: 👨‍🍳

jchris:
looks great! I will have to check to see that DocBody is backwards compatible with old apps but I'm guessing yes. Merging now.

PR #65, opened 2024-01-10
re-enable memory build

jchris:
currently failing:

```
Building dist/memory/fireproof.esm.js
✘ [ERROR] Could not resolve "fs"

    ../encrypted-blockstore/dist/node/store-memory.esm.js:3490:15:
      3490 │ import fs from "fs";
           ╵                ~~~~

  The package "fs" wasn't found on the file system but is built into node. Are you
  trying to bundle for node? You can use "platform: 'node'" to do that, which will
  remove this error.

✘ [ERROR] Could not resolve "crypto"

    ../encrypted-blockstore/dist/node/store-memory.esm.js:6487:20:
      6487 │ import crypto2 from "crypto";
           ╵                     ~~~~~~~~

  The package "crypto" wasn't found on the file system but is built into node. Are
  you trying to bundle for node? You can use "platform: 'node'" to do that, which
  will remove this error.

Error dist/memory/fireproof.esm.js
```


jchris:
@thedanchez thanks for including this in your tsup work

PR #66, opened 2024-01-12
refactor: adds tsconfig alias path in core for @fireproof/encrypted-blockstore

thedanchez:
Minor refactor PR adding the alias `path` to get rid of the `@ts-ignore` comments when importing from `@fireproof/encrypted-blockstore`


thedanchez:
This change will go in as part of https://github.com/fireproof-storage/fireproof/pull/61

PR #67, opened 2024-01-12
0.17 w pail batch put

jchris:
new pail uses


packages/fireproof/test/crdt.test.js
```diff
@@ -177,7 +177,7 @@ describe('Compact a named CRDT with writes', function () {
     for await (const blk of crdt.blockstore.entries()) {
       blz.push(blk)
     }
-    equals(blz.length, 17)
+    equals(blz.length, 35)
```
jchris: this was the only change I had to make to the tests

jchris:
implementing batch and now I'm uncovering more type details. Now is a good time to deal with stuff but I might need help.
Screenshot 2024-01-12 at 1 23 50 PM

PR #74, opened 2024-01-20
build: use tsup to bundle use-fireproof

thedanchez:
The migration to `tsup` for bundling continues -- next on the list is the React package!


thedanchez:
The React package is feeling clean, tidy and stable now -- I took the liberty of bringing over the JSDoc comments I had for the SolidJS package over here as part of this migration and cleanup.

- [x] `tsup` builds successfully
- [x] Migrated tests from Jest to Vitest and tests now pass
- [x] Code formatted using Prettier
- [x] ESLint config updated and linting is now working 
- [x] Cleaned up `devDependencies` in `package.json`

packages/fireproof/src/fireproof.ts
```diff
@@ -1,5 +1,2 @@
-// ts-unused-exports:disable-next-line
 export * from './database'
-
-
-
+export * from "./types"
```
jchris, line 2: is this the thing you said you replaced with something automatic?

jchris (approved):
thanks

PR #80, opened 2024-01-24
fix: stabilizes inferred generic typings

thedanchez:
Discovered an issue while kicking the tires with the SolidJS adapter where the TypeScript compiler was not inferring the correct typings for object fields. Came to realize that our `DocRecord` type needed to be refined further. This should stabilize the types and give better autocomplete.

Concrete example is, if you had something like `{ foo: true }` that was passed into the generics formula, the compiler would say your object had a field `foo` whose available values were `DocFragment & true` instead of simply resolving to `boolean` which is a valid `DocFragment` value type.

### DIFF
--- a/packages/fireproof/src/types.ts
+++ b/packages/fireproof/src/types.ts
@@ -25,14 +25,19 @@ export type DocFragment =
   | DocFragment[]
   | { [key: string]: DocFragment }
 
-export type DocRecord = Record<string, DocFragment>;
-export type Doc<T extends DocRecord = {}> = DocBase & DocBody<T>
+export type DocRecord<T> = {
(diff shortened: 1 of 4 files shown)


packages/fireproof/src/types.ts
```diff
@@ -25,15 +25,20 @@ export type DocFragment =
   | DocFragment[]
   | { [key: string]: DocFragment }
 
-export type DocRecord = Record<string, DocFragment>;
-export type Doc<T extends DocRecord = {}> = DocBase & DocBody<T>
+export type DocRecord<T> = {
+  [K in keyof T]: DocFragment;
+};
```
thedanchez, line 30: This is the TypeScript sauce we need

PR #83, opened 2024-01-30
Use connectors for remote storage

jchris:
This will allow you to run Fireproof in edge environments with no disk or local storage, by using a connector for environment-specific kv storage.

This will enable access to the database from lambda and edge functions.

### DIFF
--- a/packages/encrypted-blockstore/src/loader.ts
+++ b/packages/encrypted-blockstore/src/loader.ts
@@ -85,10 +85,8 @@ export class Loader implements Loadable {
       this.metaStore = ebOpts.store.makeMetaStore(this)
       if (!this.metaStore || !this.carStore || !this.remoteWAL)
         throw new Error('stores not initialized')
-      console.log('loader load')
       const metas = this.ebOpts.meta ? [this.ebOpts.meta] : await this.metaStore!.load('main')
       if (metas) {
-        console.log('loader load metas', metas.length)
         await this.handleDbMetasFromStore(metas)
       }
     })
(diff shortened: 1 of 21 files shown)


packages/encrypted-blockstore/src/store-node.ts
```diff
@@ -6,12 +6,14 @@ import { mkdir, readFile, writeFile, unlink } from 'fs/promises'
 import type { AnyBlock, AnyLink, DbMeta } from './types'
 import { STORAGE_VERSION, MetaStore as MetaStoreBase, DataStore as DataStoreBase } from './store'
 import { RemoteWAL as RemoteWALBase, WALState } from './remote-wal'
+import type { Loader } from './loader'
+
+export const makeDataStore = (name: string) => new DataStore(name);
+export const makeMetaStore = (name: string) => new MetaStore(name);
+export const makeRemoteWAL = (loader: Loader) => new RemoteWAL(loader);
+
```
jchris, line 14: an example of the factory interface

packages/fireproof/src/crdt.ts
```diff
@@ -53,8 +53,8 @@ export class CRDT {
         return { head: this.clock.head } as TransactionMeta
       },
       autoCompact: this.opts.autoCompact || 100,
-      crypto,
-      store,
+      crypto: this.opts.crypto || crypto,
+      store: this.opts.store || store,
```
jchris, line 57: this part deserves documentation

packages/encrypted-blockstore/src/loader.ts
```diff
@@ -433,7 +444,45 @@ export class Loader {
   }
 }
 
+// duplicated in @fireproof/connect
```
jchris, line 447: maybe this means the modules should be one module? not making that change now. Maybe there are other ways to clean up this type dependency graph

Issue #81, opened 2024-01-29
Typescript Generics in react-hooks


davidjamesb:
First of all great project and I'm excited for the work ongoing in this space and this project in particular!

I'm following the vite/react tutorial [here](https://use-fireproof.com/docs/react-tutorial/) and ran into issues with the typescript compiler:

![image](https://github.com/fireproof-storage/fireproof/assets/2739322/28677686-9362-4d41-9bbb-affb3fa69aa5)

Switching to an interface and extending `DocBase` gets past this hurdle:

![image](https://github.com/fireproof-storage/fireproof/assets/2739322/c589972c-af33-404c-bd9d-5026afffcf8e)

However, the types returned by the `useDocument` hook are inferred as `DocBase` leading to a lot of casting when using the `todo` object as an actual instance of `ToDo` later on in the react component.

I expected there to be a version such as `useDocument<T>(...)` and looking through the source for `useFireProof.tsx` there does appear to be comments alluding to this, but such signature doesn't appear to exist. I understand there has been some recent work regarding generics in #80 so maybe this is due to come. Either that or I'm using it wrong.
![image](https://github.com/fireproof-storage/fireproof/assets/2739322/cb205d33-5312-4092-87bc-6d73e159eb92)

The siutation is the same for `useLiveQuery`.

Finally, a tangent but related - whilst digging around the source I noticed that we have generic signatures for most functions on `database.ts` except `subscribe`. It's my understanding that this function will listen for all changes that occurred in the database since the latest clock sync with which documents of multiple types could have been changed.

Whilst working with other document databases, the common solution is to add a `type : string` field to each document and have a discriminator function map from the serialized JSON to the correct type at runtime. Would this be the recommendation when using fireproof e.g. handle it in the `(changes) => {...}` callback or is there scope to 'bake' this into the source by embedding the object type information into the JSON or providing an overload e.g:

```
type DocDiscriminator = (doc: DocBase) => boolean;

subscribe(listener: ListenerFn | NoUpdateListenerFn, updates?: boolean, discriminator?: DocDicriminator): () => void {
}
```

thedanchez:
Ah, thanks for raising this! Was an oversight on my part. Got missed as I was porting over the work from the SolidJS adapter.

thedanchez:
@davidjamesb just merged in the PR to fix this issue. Should have a new version of the React package published soon!

PR #85, opened 2024-01-31
fix: updates React hooks to support generics properly

thedanchez:
This PR is in response to the issue posted here: https://github.com/fireproof-storage/fireproof/issues/81

Was an oversight on my part

## Remaining TODOs
- Kick the tires in the React example app


jchris (approved):
This looks great but I haven't run it locally. I'll wait to merge until I have time or get a +1 that it's ready.

thedanchez:
@jchris sounds good -- I also want to kick the tires a little bit and maybe take the opportunity to start off a new local react example app so that we can eventually deprecate the other ones and just build on the new one if that's cool with you?

thedanchez:
Converting this to draft because did some tests and saw some quirks to iron out

thedanchez:
@jchris so I made a slight change to the `useDocument` API interface for the React SDK. Ended up going down this path after seeing undesirable behavior while testing on the example UI with how our internal `useCallback` and `useEffect` kept on being re-calculated or re-ran.

The change is, instead of passing a raw anonymous object, the user will instead pass in a function that generates the document. 

```tsx
// Before -- raw object input
const [...] = useDocument({ text: "", date: new Date(), completed: false });

// After -- generator function input
const [...] = useDocument(() => ({ text: "", date: new Date(), completed: false }));
```

In order to stabilize things while keeping things as they were, the user would have to memoize the object they are passing in which I think is not something we probably want to force on them. Looking at the above example you could see how this can happen when we have something like that `new Date()` in the object -- it returns a new value every time the React re-render cycle kicks in. And because it changes on every render, our internal implementation things that depend on that `initialDoc` will trigger their recalculations which is pretty volatile.

By making it a generator function, we can memoize it internally (taking that responsibility from the user) and it'll be static despite containing something like `new Date()` or any other dynamic value (like some value coming from component props). 

One other note here is the upcoming SolidJS SDK will go down that same path as Solid only renders once and we do not want to create "stale" versions of the initial doc object (because `new Date()` will have been evaluated only once in Solid land). Making it a generator function will allow us to grab a fresh `new Date()` value upon invocation.

Feel free to disagree with me on this change or if you think I'm off the mark here. Also, I would love for you to test things out if you're able to. Definitely would feel more comfortable knowing you were able to kick the tires on this.

I made a new React example app (under `examples/react`) as part of this that you can run from the root project directory using `pnpm start:react`.

I recommend running two shells, one which runs `pnpm build:watch:react` and the other will run `pnpm start:react`

jchris:
this sounds correct, looks correct, and works great here.

packages/react/src/useFireproof.tsx
```diff
@@ -91,74 +108,99 @@ export const FireproofCtx = {} as FireproofCtxValue;
...
 
     useEffect(() => {
-      refreshDoc();
-    }, []);
-
-    return [
-      doc,
-      (newDoc, replace) => {
-        if (newDoc) return replace ? setDoc(newDoc) : setDoc((d) => ({ ...d, ...newDoc }));
-        else return setDoc(initialDoc);
```
jchris, line 129: this behavior is missing in the new one

PR #88, opened 2024-02-06
build: refines github workflow

thedanchez:
Refines the repo GitHub workflows to include both lint and format checks going forward. Right now, only the React and Solid packages have clean ESLint and Prettier setups. We'll need to go down the package list and do the same for the others so we can globally enforce linting and formatting conventions for the repo


jchris:
thanks!

PR #99, opened 2024-03-05
cleanup get block function

jchris:
this is intended to be a pure refactor


jchris:
Video of the coding session is available here: https://youtu.be/GX8gCQ17kNE

jchris:
this doesn't work right in compact mode

PR #101, opened 2024-03-07
changes to support websocket connection

dhruvsoni1802:


packages/connect/src/index.ts
```diff
@@ -30,15 +30,19 @@ class ConnectRaw extends Connection {
 }
 
 export const connect = {
-  s3free: ({ blockstore }: Connectable) => {
-    const upload = 'https://udvtu5wy39.execute-api.us-east-2.amazonaws.com/uploads'
-    const download = 'https://crdt-s3uploadbucket-dcjyurxwxmba.s3.us-east-2.amazonaws.com'
-    const connection = new ConnectS3(upload, download)
+  s3free: ({ blockstore }: Connectable, databasename: string) => {
```
jchris: instead of changing existing ones for now, lets make an awsFree that is like s3Free plus ws

packages/connect/src/connect-s3.ts
```diff
@@ -58,6 +66,41 @@ export class ConnectS3 extends Connection {
     return bytes
   }
 
+  async onConnect() {
+    console.log("Is the onconnect function being called?")
```
jchris: ?

packages/connect/src/connect-s3.ts
```diff
@@ -41,8 +48,7 @@ export class ConnectS3 extends Connection {
       body: JSON.stringify(crdtEntry),
     })
     const result = await done.json()
-    if (result.status != 201)
-    {
+    if (result.status != 201) {
```
jchris: good to remove reformat-only changes for something like this

packages/connect/src/index.ts
jchris: this is the main one to fix

packages/connect/src/connect-s3.ts
dhruvsoni1802: For every file I always do a format document and then save it. However these stuff keeps happening

PR #102, opened 2024-03-22
Pail update

dhruvsoni1802:


.npmrc
```diff
@@ -1 +1 @@
-auto-install-peers=false
+auto-install-peers=true
```
jchris, line 1: I'm not sure why this was like that, we should look at history

PR #107, opened 2024-04-12
React Native - attempt #2

boorad:
This PR adds the required bits to have a fully-functional Fireproof database in a React Native mobile application.  The biggest difference is polyfilling certain global objects that are expected to be present in Node or a browser with alternatives that work in React Native.

We also add an example RN app showcasing Fireproof as the live database.

- [x] Fix [`react-native-quick-crypto`](https://github.com/margelo/react-native-quick-crypto) library
- [x] Add required polyfills
- [x] Implement `store-native` logic and use [MMKV](https://github.com/mrousavy/react-native-mmkv) as local storage
- [ ] Test in RN Fireproof example application
   - [x] Create
   - [x] Read
   - [x] Update
   - [x] Delete
   - [x] Subscribe to changes
   - [ ] Get `useLiveQuery` to work, instead of `db.allDocs()`?
   - [ ] Get `FlatList` component to not throw React Hooks error?
- [ ] Hook up and test some form of cloud sync

----

Note: Until `react-native-quick-crypto` gets support for `subtle.encrypt()` and `subtle.decrypt()`, you need to use `{public: true}` database option:

    const { database, useDocument } = useFireproof('TodoDB', {public: true});


packages/encrypted-blockstore/src/encrypt-helpers.ts
```diff
@@ -143,7 +143,7 @@ export async function encryptedEncodeCarFile(
   rootCid: AnyLink,
   t: CarMakeable
 ): Promise<AnyBlock> {
-  const encryptionKey = hexStringToUint8Array(key)
+  const encryptionKey = hexStringToArrayBuffer(key)
```
jchris: I love JavaScript

boorad:
Simulator Screenshot - iPhone 15 Pro - 2024-05-13 at 19 06 32

jchris:
The `(there's never any changes in db.subscribe())` is something that happens to a lot of people. I think I should update the API so that it doesn't pass any data, ever. Then people would know it is just an event they can use to know when to query the database. That's how I use it 90% of the time. Hence, it has an optional second argument when you pass `true` that returns the data items. The reason why is in conditions of post merge (remote events), fetching the data items can be somewhat expensive, which makes no sense when they are discarded 90% of the time. Curious for your thoughts on the API.

It would be interesting to see what issues show up when trying `useLiveQuery` as that should be the main API in react. It's probably worth bridging the gap unless there is something weird there.

boorad:
### `db.subscribe()`

Aha, so all your docs have to do is compensate for me being an idiot and not reading / understanding.  I do think you should add a bit about that boolean flag in the docs, like: 

> Note: Use `returnChangeData: boolean` argument to return the changed data

The latest commit has this working and refreshing the Todo List after any change, but it is kind of slow.  Like a second or two of logging, and then the mobile UI updates.  That could be me not exactly getting the call semantics correct.

boorad:
### `useLiveQuery()`
Can you talk to how the index is stored (prolly trees)?  From the docs:

> By default, Fireproof stores data in the browser's local storage

RN doesn't have local storage, so wondering how I can override and store in MMKV or similar.

I see the proper number of documents coming from `useLiveQuery` but they're all undefined, and have that `{_key: ..., _value: ...} format.

boorad:
closing in favor of an external storage module: https://github.com/boorad/fireproof-react-native

PR #115, opened 2024-05-28
feat: get UCAN connector working

travis:
introduce a new, "v2" connector and leave the old one in place - we may want to get rid of that too, soon?


examples/next-app/src/app/page.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { connect } from "@fireproof/ipfs"
+import { useState } from "react";
+import { useForm } from "react-hook-form";
+
+type AuthFormValues = {
+  email: string
+}
+
+export default function Home () {
+  const [db] = useState<Database>(fireproof('example'))
```
jchris, line 15: I tried to make it unnecessary to wrangle details like this for users of the React hooks. See eg this where the `id` is the database name, taken from the url path slug. I'm not sure it's right, but wanna make sure you grok that approach early on and give me feedback: https://github.com/fireproof-storage/firehouse-chat/blob/bc8cf00df8cb242456ddef3ffd84bea24faf7fe3/src/components/Channel.tsx#L44

packages/connect-ipfs/package.json
```diff
@@ -62,8 +44,8 @@
     "@ucanto/interface": "^10.0.1",
     "@web3-storage/clock": "^0.4.1",
     "@web3-storage/pail": "^0.6.0",
-    "@web3-storage/w3up-client": "^12.5.2",
+    "@web3-storage/w3up-client": "^13.1.1",
```
jchris, line 65: 🚀

packages/connect-ipfs/src/connect-ipfs-account.ts
```diff
@@ -111,56 +119,44 @@ export class AccountConnectIPFS extends DatabaseConnectIPFS {
...
     await this.authorizing
     return this.client!
   }
 
-  clockSpaceDIDForDb() {
+  clockSpaceDIDForDb () {
     return this.client!.currentSpace()!.did()
   }
 
-  async connectClient() {
```
jchris, line 131: glad to see this go :)

packages/encrypted-blockstore/package.json
```diff
@@ -3,55 +3,28 @@
...
-      "default": "./dist/lib/store-node.js"
+      "import": "./src/store-node.ts"
     },
     "./store-web": {
-      "import": "./dist/lib/store-web.js",
-      "require": "./dist/lib/store-web.cjs",
-      "types": "./dist/lib/store-web.d.ts",
-      "script": "./dist/lib/store-web.global.js",
-      "default": "./dist/lib/store-web.js"
+      "import": "./src/store-web.ts"
```
jchris: 🫡

packages/encrypted-blockstore/src/store-remote.ts
```diff
@@ -116,16 +116,22 @@ export class RemoteMetaStore extends MetaStoreBase {
     return this.handleByteHeads(byteHeads, branch)
   }
 
-  async save(meta: DbMeta, branch: string = 'main') {
+  async save (meta: DbMeta, branch: string = 'main') {
+    console.log("SAVING")
     const bytes = new TextEncoder().encode(this.makeHeader(meta))
```
jchris: I think this whole thing might be double encoded, leading to some of the issues. Maybe something to form an opinion on about over time.

examples/next-app/src/app/page.tsx
travis: yea - just didn't want to deal with one more layer while I'm debugging! I think having a react hook for most usage is the right call

packages/encrypted-blockstore/package.json
travis: haha, this might be short lived! just makes debugging easier

packages/encrypted-blockstore/src/store-remote.ts
travis: ah that could be the issue - let me try it...

jchris:
Screenshot 2024-06-05 at 11 10 24 AM

it connects to the same space, and requests access to the correct space, but doesn't seem to sync.

examples/next-app/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@fireproof/core": "workspace:^",
+    "@fireproof/ucan": "workspace:^",
```
jchris, line 13: I did the full rename to UCAN

jchris:
Screenshot 2024-06-13 at 1 50 56 PM

next app isn't loading for some reason, complains about the route

travis:
changes lgtm! I'll take a look at the next app...

PR #118, opened 2024-06-10
initial sql commit

mabels:
This is the implementation of a sql connect method for fireproof.


packages/connect-sql/src/connect-sql-node.ts
```diff
@@ -0,0 +1,132 @@
...
+    console.log('sql onConnect-3')
+    this.metaStore = await MetaStoreFactory(this.dbConn)
+    console.log('sql onConnect-4')
+    return Promise.resolve()
+  }
+
+  /**
+   * metaDownload Function
+   *
+   * This function downloads metadata for a specific name and branch.
```
jchris: I scrubbed this comment from the original. I think it might be wrong / confusing, so best to remove here also

packages/connect-sql/src/data-type.ts
```diff
@@ -0,0 +1,81 @@
...
+export class DataSQLRecordBuilder {
+    #car?: string;
+    #size?: string;
+    #data?: Uint8Array;
+
+    static fromUploadParams(data: Uint8Array, params: UploadDataFnParams): DataSQLRecordBuilder {
+        const bld = new DataSQLRecordBuilder()
+        bld.#car = params.car
+        bld.#data = data
+        bld.#size = params.size
```
jchris: I think we don't use`size` anywhere so perhaps worth removing if that's true. It's purpose is ostensibly as a gaurd against the s3 signed upload URL handler accepting DOS sized files, but I don't see it getting used there. Maybe worth adding there if possible, otherwise worth removing from the entire `UploadDataFnParams` type across connectors. This is a good time in the release lifecycle to do these kinds of global changes.

jchris: https://github.com/fireproof-storage/valid-cid-s3-bucket/blob/0f8f8d44f11cd4bd8b7665316d6676b49014830e/getSignedURL/app.ts#L220

packages/connect-sql/src/meta-type.ts
```diff
@@ -0,0 +1,91 @@
...
+    insertStmt?: Statement;
+    selectStmt?: Statement;
+    readonly table: string;
+    constructor(client: Database, table: string = 'metastore') {
+        this.client = client
+        this.table = table
+    }
+    async start(): Promise<SQLiteMetaStore> {
+        await this.client.prepare(`CREATE TABLE IF NOT EXISTS ${this.table} (
+        name TEXT not null,
```
jchris, line 83: we track the CID and parent CIDs of the blob. and when a blob comes in with a parent CID that matches a current blob's CID, it replaces that current blob, which can be deleted. here is [the client side](https://github.com/fireproof-storage/fireproof/blob/main/packages/connect-netlify/src/connect-netlify.ts#L38) and [here is the server](https://github.com/fireproof-storage/fireproof/blob/82daf5a7cbe8d2b37742d35a7d52591db3a27194/packages/connect-netlify/src/server.ts#L29)

packages/encrypted-blockstore/src/connection.ts
```diff
@@ -48,17 +48,19 @@ export abstract class Connection {
...
   }
 
-  connectMeta({ loader }: { loader: Loader | undefined}) {
-    if (!loader) throw new Error('loader is required')
+  async connectMeta({ loader }: { loader: Loader | undefined }) {
+    if (!loader) throw new Error('connectMeta: loader is required')
     this.loader = loader
     this.taskManager = new TaskManager(loader)
-    this.onConnect()
+    await this.onConnect()
```
jchris, line 63: lets watch ths for does it change expected behavior in other connectors. I think there are places that expect this to be syncronous.

jchris (commented):
thanks for kicking this off!

jchris:
this is using SQL as an external connector. it might be worth exploring what happens if you write this in a completely different part of the code base. eg instead of another connector (for background sync) it could be a storage option (for local compute/store) which I think is a better fit. Your code here looks good it's just I'm not sure it's solving the problem people want to use the SQL adapter for.

I would take a look at [this node/fs storage module](https://github.com/fireproof-storage/fireproof/blob/main/packages/encrypted-blockstore/src/store-node.ts), and how to make a copy of it that uses SQL. You'll also need to [wrench a config into the core module like this](https://github.com/fireproof-storage/fireproof/blob/main/packages/fireproof/src/eb-node.ts) and the [corresponding build here](https://github.com/fireproof-storage/fireproof/blob/82daf5a7cbe8d2b37742d35a7d52591db3a27194/packages/fireproof/tsup.config.ts#L28)

In the end the approach I'm describing here would show up as:

```ts
import { fireproof } from "@fireproof/core/sql"
```

Which I know is kinda garnly so probably looking at how that storage system plumbed will give you some other ideas about cleanup.

packages/connect-sql/src/data-type.ts
mabels: i think we should not mix up things here --- this is about to create a sql adapter --> and your approach is totally fine but should be done in a future PR.

packages/connect-sql/src/meta-type.ts
mabels: thx for the explaining it -- but this line seams not related to this.

packages/encrypted-blockstore/src/connection.ts
mabels: we have to change it in any ways, i'm aware it could breaks things -- but it's impossible to sync a async connect function in javascript so the await is need and no problem on the first sight the onConnect is also async, and there is no problem if onConnect does not return a Promise modern javascript can handle that anyways

mabels:
> this is using SQL as an external connector. it might be worth exploring what happens if you write this in a completely different part of the code base. eg instead of another connector (for background sync) it could be a storage option (for local compute/store) which I think is a better fit. Your code here looks good it's just I'm not sure it's solving the problem people want to use the SQL adapter for.
> 
> I would take a look at [this node/fs storage module](https://github.com/fireproof-storage/fireproof/blob/main/packages/encrypted-blockstore/src/store-node.ts), and how to make a copy of it that uses SQL. You'll also need to [wrench a config into the core module like this](https://github.com/fireproof-storage/fireproof/blob/main/packages/fireproof/src/eb-node.ts) and the [corresponding build here](https://github.com/fireproof-storage/fireproof/blob/82daf5a7cbe8d2b37742d35a7d52591db3a27194/packages/fireproof/tsup.config.ts#L28)
> 
> In the end the approach I'm describing here would show up as:
> 
> ```ts
> import { fireproof } from "@fireproof/core/sql"
> ```
> 
> Which I know is kinda garnly so probably looking at how that storage system plumbed will give you some other ideas about cleanup.

what you described is where i'm headed to but as said before this is not done yet!

packages/connect-sql/src/meta-type.ts
jchris: I think the table need a column for cid and a column for parents (short array of cids, don't need anything fancy)

jchris: but that is for a connector, not a storage, to track multi-writer

jchris:
This is super exciting -- will be interesting to see how it compares to node/fs on the benchmark: https://github.com/dstanesc/fireproof-playground

PR #119, opened 2024-06-11
Storage engine splitting 

dhruvsoni1802:


mabels:
Hi,
i tried to review this but the title assumes that there should be some "functional" changes but all what i saw it formatting.
I recommend that we split formatting PR's from "functional" ones. I know that's hard in the process but to make reviews matter it's somehow needed on my opinion.

Meno

packages/encrypted-blockstore/src/loader.ts
```diff
@@ -241,19 +269,29 @@ export class Loader implements Loadable {
...
     // Maximum size of each CAR file as 1mb?
-    const { cid, bytes } = await this.prepareCarFile(roots[0], t, !!opts.public)
-    await this.carStore!.save({ cid, bytes })
-    await this.cacheTransaction(t)
-    const newDbMeta = { cars: [cid], key: this.key || null } as DbMeta
-    await this.remoteWAL!.enqueue(newDbMeta, opts)
-    await this.metaStore!.save(newDbMeta)
-    await this.updateCarLog([cid], fp, !!opts.compact)
-    return [cid]
+    const cars = await this.prepareCarFiles(roots, t, !!opts.public)
```
jchris: the big move here is having the return type of this function potentially include multiple car files instead of one. this is a follow up to a previously merged pr that changes the header format to accomodate split file outputs.

packages/encrypted-blockstore/src/loader.ts
```diff
@@ -241,19 +269,29 @@ export class Loader implements Loadable {
...
+    const cids: AnyLink[] = []
+    for (const car of cars) {
+      const { cid, bytes } = car
+      await this.carStore!.save({ cid, bytes })
+      await this.cacheTransaction(t)
```
jchris: `cacheTransaction` should only happen once at the end of the loop

packages/encrypted-blockstore/src/loader.ts
```diff
@@ -241,19 +269,29 @@ export class Loader implements Loadable {
...
+      const newDbMeta = { cars: [cid], key: this.key || null } as DbMeta
+      await this.remoteWAL!.enqueue(newDbMeta, opts)
+      await this.metaStore!.save(newDbMeta)
+      await this.updateCarLog([cid], fp, !!opts.compact)
+      cids.push(cid)
```
jchris, line 283: same with the other lines:

```
const newDbMeta = { cars: [cid], key: this.key || null } as DbMeta
       await this.remoteWAL!.enqueue(newDbMeta, opts)
       await this.metaStore!.save(newDbMeta)
       await this.updateCarLog([cid], fp, !!opts.compact)
```

they should happen after the loop ends and the cids are pushed.

packages/encrypted-blockstore/src/loader.ts
```diff
@@ -264,15 +302,54 @@ export class Loader implements Loadable {
...
+          theKey && this.ebOpts.crypto
+            ? await encryptedEncodeCarFile(
+                this.ebOpts.crypto,
+                theKey,
+                roots[0],
+                clonedt
+              )
+            : await encodeCarFile([roots[0]], clonedt)
+        carFiles.push(carFile)
+        clonedt = new CarTransaction(t.parent)
```
jchris: i think we need to reset newcount to 0 here

jchris:
> Hi, i tried to review this but the title assumes that there should be some "functional" changes but all what i saw it formatting. I recommend that we split formatting PR's from "functional" ones. I know that's hard in the process but to make reviews matter it's somehow needed on my opinion.
> 
> Meno

It would be awesome to standardize the vscode env and build chain so we don't have to worry about format shifting

vmx:
> It would be awesome to standardize the vscode env

Using some automated formatter, independent of the edior/IDE would be great. Something like [dprint](https://dprint.dev/) or [prettier](https://prettier.io/).

packages/encrypted-blockstore/src/loader.ts
```diff
@@ -264,15 +302,55 @@ export class Loader implements Loadable {
...
     isPublic: boolean
-  ): Promise<{ cid: AnyLink; bytes: Uint8Array }> {
+  ): Promise<{ cid: AnyLink; bytes: Uint8Array }[]> {
     const theKey = isPublic ? null : await this._getKey()
-    return theKey && this.ebOpts.crypto
-      ? await encryptedEncodeCarFile(this.ebOpts.crypto, theKey, root, t)
-      : await encodeCarFile([root], t)
+    const carFiles: { cid: AnyLink; bytes: Uint8Array }[] = []
+    let size = 0
+    let threshold = 1024 * 1024
```
jchris: This threshold should be configrable via `this.ebOpts`

packages/encrypted-blockstore/src/transaction.ts
```diff
@@ -191,4 +191,5 @@ export type BlockstoreOpts = {
   public?: boolean
   meta?: DbMeta
   name?: string
+  threshold?:number
```
jchris: formatter will fix this #121

jchris:
Screenshot 2024-06-12 at 7 07 38 PM
 was able to crash under stress... still investigating before merge

PR #126, opened 2024-06-16
formatted and restructured

mabels:
Explore the possiblities to restructure the project.


src/crdt-clock.ts
```diff
@@ -1,18 +1,18 @@
 import { clockChangesSince } from "./crdt-helpers";
 import type { EncryptedBlockstore, CarTransaction } from "./storage-engine/";
-import type { DocUpdate, ClockHead } from "./types";
+import type { DocUpdate, ClockHead, IndexKeyType } from "./types";
 import { advance } from "@web3-storage/pail/clock";
 import { root } from "@web3-storage/pail/crdt";
 import { applyHeadQueue, ApplyHeadQueue } from "./apply-head-queue";
 
-export class CRDTClock {
+export class CRDTClock<T, K extends IndexKeyType> {
```
jchris: I think this CRDT stuff should be decoupled from the DbIndex stuff that uses the index keys. Maybe there is another type here? Or maybe I'm missing something. This type might not be one that needs extensibility.

jchris (commented):
I like the formatting fixups

tests/fireproof/indexer.test.ts
```diff
@@ -256,7 +256,7 @@ describe("basic Index upon cold start", function () {
     assert(result2);
     equals(result2.rows.length, 3);
     equals(didMap, 0);
-    await crdt2.bulk([{ key: "abc4", value: { title: "despicable", score: 0 } }]);
+    await crdt2.bulk([{ id: "abc4", value: { title: "despicable", score: 0 } }]);
```
jchris, line 274: note we are changing a public signature here. a minor one so its ok.

jchris: I guess it's only public on the read side.

scripts/solid-js/README.md
```diff
@@ -83,7 +83,7 @@ Create (or modify) a document in your Fireproof database.
 
 ```ts
 // API Signature
-function createDocument<T extends DocRecord<T>>(initialDocFn: Accessor<Doc<T>>): CreateDocumentResult<T>;
+function createDocument<T extends DocTypes>(initialDocFn: Accessor<Doc<T>>): CreateDocumentResult<T>;
```
jchris, line 86: we should do a pass before merge to look at all the public changes and consider which ones make sense to un-churn. I think this is better but I wanna see how they stack up in aggregate. 

Maybe we have the improvements for the next release to make eval easier.

src/connectors/connect-aws/connect-s3.ts
```diff
@@ -59,14 +59,16 @@ export class ConnectS3 extends Connection {
       throw new Error("failed to upload data " + JSON.parse(result.body).message);
     }
     this.parents = [event.cid];
-    return null;
+    return undefined
```
jchris: why the change? I have been wrapping my head around null vs undefined with no concrete answers. This is a public signature, in the sense that all connector authors must deal with it.

src/database.ts
```diff
@@ -109,49 +115,48 @@ export class Database implements Connectable {
...
       return () => {
-        this._noupdate_listeners.delete(listener);
+        this._noupdate_listeners.delete(listener as ListenerFn<{}>);
       };
     }
   }
 
   // todo if we add this onto dbs in fireproof.ts then we can make index.ts a separate package
-  async query<K extends IndexKeyType, T>(
+  async query<K extends IndexKeyType, T extends DocTypes, R extends DocFragment = T>(
```
jchris, line 140: wanna think about how this feels in apps

src/indexer-helpers.ts
```diff
@@ -168,15 +179,11 @@ export async function applyQuery<T, K extends IndexKeyType>(
...
-      if (row.row && !row.value) {
-        row.value = row.row;
-        delete row.row;
-      }
-      return row;
-    }),
+    rows: resp.result.map((row) => ({
+      id: row.id,
+      key: charwise.decode(row.key),
+      value: row.row,
```
jchris, line 185: I wonder why the old code has the conditional mutation. applyQuery used to consume some other types (I think we cleaned up prolly-trees's return structure at some point, the row.row = row.value might have been to handle multiple versions of prolly-tree with this same function

src/indexer-helpers.ts
```diff
@@ -188,13 +195,19 @@ export function encodeKey(key: DocFragment): string {
...
 // ProllyNode type based on the ProllyNode from 'prolly-trees/base'
-interface ProllyNode<K extends IndexKeyType, T> extends BaseNode {
-  getAllEntries(): PromiseLike<{ [x: string]: any; result: IndexRow<K, T>[] }>;
-  getMany(removeIds: K[]): Promise<{ /* [x: K]: unknown; */ result: IndexKey<K>[] }>;
-  range(a: IndexKey<K>, b: IndexKey<K>): Promise<{ result: IndexRow<K, T>[] }>;
-  get(key: string): Promise<{ result: IndexRow<K, T>[] }>;
-  bulk(bulk: IndexUpdate<K>[]): PromiseLike<{
+interface ProllyNode<K extends IndexKeyType, T extends DocFragment> extends BaseNode {
+  getAllEntries(): PromiseLike<{ [x: string]: any; result: ProllyIndexRow<K, T>[] }>;
+  getMany<KI extends IndexKeyType>(removeIds: KI[]): Promise<{ /* [x: K]: unknown; */ result: IndexKey<K>[] }>;
```
jchris, line 207: these keys are all strings, post charwise encoding

src/indexer.ts
```diff
@@ -0,0 +1,263 @@
...
+    //     // @ts-ignore
+    //     if (header.head) throw new Error('cannot have head in idx header')
+    //     if (header.indexes === undefined) throw new Error('missing indexes in idx header')
+    //     // for (const [name, idx] of Object.entries(header.indexes)) {
+    //     //   index({ _crdt: crdt }, name, undefined, idx as IdxMeta)
+    //     // }
+    //   })
+  }
+
+  applyMapFn(name: string, mapFn?: MapFn<T>, meta?: IdxMeta) {
```
jchris, line 89: once this function has full test coverage I think we can make it much shorter

src/storage-engine/encrypt-codec.ts
```diff
@@ -1,7 +1,20 @@
...
+}
+
+interface DecryptOpts {
+  readonly key: ArrayBuffer;
+  readonly value: {
+    readonly bytes: Uint8Array;
+    readonly iv: Uint8Array
+  };
+}
+
```
jchris, line 17: `:high_five:`

src/storage-engine/loader-helpers.ts
```diff
@@ -46,8 +42,10 @@ export async function parseCarFile(reader: CarReader): Promise<CarHeader> {
   const header = await reader.get(roots[0]);
   if (!header) throw new Error("missing header block");
   const { value } = await decode({ bytes: header.bytes, hasher, codec });
-  // @ts-ignore
-  if (value && value.fp === undefined) throw new Error("missing fp");
-  const { fp } = value as { fp: CarHeader };
-  return fp;
+  const fpvalue = value as { readonly fp: CarHeader };
+  // @jchris where is the fp attribute coming from?
```
jchris, line 46: the `CarHeader` type name is misleading -- it's not a CAR file thing, it's just the data structure we put in the "root" block of the CAR file. Maybe it should be called `FireproofHeader`.

comes from 

```
export async function encodeCarHeader(fp: CarHeader) {
  return (await encode({
    value: { fp },
    hasher,
    codec
  })) as AnyBlock
}
```

the purpose is to make it clear this is a FireproofHeader, and not some other applications, when we load it from the storage engine.

src/storage-engine/loader.ts
```diff
@@ -324,7 +324,7 @@ export class Loader implements Loadable {
         files.push(meta.cid as AnyLink);
       }
     }
-    return { files };
+    return { ...result, files };
```
jchris, line 322: this one might be spooky, maybe need to console.log what comes in

src/version.ts
```diff
@@ -1 +1 @@
-export const PACKAGE_VERSION = "0.18.0";
+export const PACKAGE_VERSION = "dev";
```
jchris, line 1: this will get copied from package.json by scripts, and then used for the file-path for saving stuff, so maybe a strategy that yields eg 0.19-dev in situations like this would be good for sanity long term

mabels:
it's moved to the main repo

PR #131, opened 2024-07-02
bring in dashboard

jchris:
it would be great if this imported `use-fireproof` via workspace, but I can't get it to work


examples/dashboard/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "preview": "vite preview",
+    "typecheck": "tsc"
+  },
+  "dependencies": {
+    "@remix-run/node": "^2.10.0",
+    "@remix-run/react": "^2.10.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
+    "react-router-dom": "^6.24.0",
+    "use-fireproof": "^0.18.0"
```
jchris, line 24: this should be workspace but it doesnt work yet

jchris: it would be cool to have help with this, @mabels

PR #132, opened 2024-07-07
New structure

mabels:


src/blockstore/loader.ts
```diff
@@ -0,0 +1,615 @@
...
+
+interface Startable {
+  start(): Promise<void>;
+}
+
+class EnsureStart<T extends Startable> {
+  readonly oncePerStore = new ResolveOnce<T>();
+
+  reset() {
+    this.oncePerStore.reset();
```
boorad, line 80: `ResolveOnce` doesn't have a `reset()` function

src/blockstore/loader.ts
```diff
@@ -0,0 +1,615 @@
...
+    const toClose = await Promise.all([this.carStore(), this.metaStore(), this.fileStore(), this.remoteWAL()]);
+    await Promise.all(
+      toClose.map((store) => {
+        store.close();
+      }),
+    );
+    this.ensureCarStore.reset();
+    this.ensureFileStore.reset();
+    this.ensureRemoteFileStore.reset();
+    this.ensureMetaStore.reset();
```
boorad: same for these `reset()` calls

boorad (commented):
found a couple of issues in loader

src/blockstore/remote-wal.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  async destroy() {
+    await this.ready();
+    await this._destroy();
+  }
+
+  abstract _load(branch?: string): Promise<WALState | Falsy>;
+  abstract _save(state: WALState, branch?: string): Promise<void>;
+  abstract _close(): Promise<void>;
+  abstract _destroy(): Promise<void>;
```
boorad, line 193: Why the underscores on these methods?

src/blockstore/loader.ts
mabels: you need to do an pnpm i

boorad: `pnpm i` was already run.  If it resolves these four, it doesn't resolve the one above on line 80.

PR #146, opened 2024-07-24
Add failing test for successive queries with functions

necrodome:
The problem seems to be with [makeName](https://github.com/fireproof-storage/fireproof/blob/df22d250ebf7cb79b5c03830424eef8cff1f17a0/src/database.ts#L245), as it is sensitive to how the function is defined.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,10 +27,10 @@ jobs:
         run: pnpm run lint
       - name: build
         run: pnpm run build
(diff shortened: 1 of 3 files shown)


jchris:
I usually run this before I commit: `pnpm i && pnpm format --write && pnpm build && pnpm test`

We should probably document something like that better :)

jchris:
test suite passes and smoke test issue is being worked on elsewhere. merging

PR #167, opened 2024-08-12
test for deterministic iv

mschoch:
test asserts that the encode result is the same, which can only happen when the same iv is used for provided input


mschoch:
@jchris not sure if you had a higher-level test in mind, but this does fail before and pass with the change.

jchris:
thanks, merged to branch

PR #169, opened 2024-08-13
Lock cement

mschoch:
CI on main keeps grabbing an incompatible version of cement


mschoch:
@mabels would prefer we not pin

mschoch:
Reopening per @jchris

jchris:
I agree with @mabels we shouldn't pin but main is red right now, so we'll unpin as soon as we can, and this will let us merge.

PR #172, opened 2024-08-14
test for concurrent compaction

jchris:
currently failing test should pass when compaction can be run concurrent to live load


tests/fireproof/fireproof.test.ts
```diff
@@ -174,21 +174,21 @@ describe("benchmarking with compaction", function () {
   beforeEach(async function () {
     // erase the existing test data
     await rt.SysContainer.start();
-    db = new Database("test-benchmark-compaction", { autoCompact: 3, public: true });
+    db = new Database("test-benchmark-compaction", { autoCompact: 3 });
   });
-  it.skip("passing: insert during compaction", async function () {
+  it("insert during compaction", async function () {
```
jchris, line 179: @mabels I removed a skip!

PR #176, opened 2024-08-15
introduce abstraction for websocket

mschoch:
access websocket from node is different from browser partykit handles browser case automatically
but requries us to pass in WebSocket from ws

currently this does not do the browser side


mschoch:
I was wrong, we don't need this.

PR #183, opened 2024-08-19
Introduction of a SuperThis

mabels:
This should change no behavior but finally bring a stable dependency injection platform to the codebase.


mschoch:
This is a pretty wide change and I would benefit from some additional context to properly review.

What is a SuperThis?  Is this a common construct in the js/ts world for dependency injection purposes?

mabels:
The superthis is the finalization of the passthrough of the logger

 --- but due to the endless loop fix I ran into the situation that the "async" process doesn't enough about the "runtime" and I'm not able to pass the runtime down.
I decided to go for the ugliest but only working pattern to do "dependency inject" pass the runtime in a single-process execution model. 
And this is to add an explicit parameter passed through all layers. I named it superthis because it is the second parameter beside the first "this/self" --- in other runtimes, it is named "context" which you have to pass and with is not typed.

mschoch:
At least we agree that it is ugly.

mschoch:
OK, so since my ability to review this in the abstract is somewhat limited.  I instead set out to test this branch with the gateway I've been working on.  It took some effort to even get things running as well as before, but it was useful in at least letting me appreciate these changes in practice (most of the questions I had could be answered by the PR itself).

Also, I've heavily modified the PK logic to function almost identically to the file gateway now (just writes to the party room instead of disk).  Eventually we'll make use of the websocket again for the realtime updates.

As I said, my test still doesn't work.  It doesn't hang, Bob just never sees the data that Alice wrote (syncing to same PK).  

One question I have, in the test, I try to force the update with (shown are my previous and current way of doing it)

```
- await bob.blockstore.loader?.remoteMetaStore?.load('main')
+ await connectionBob.loader?.remoteMetaStore?.load('main')
```

Is that correct?

mschoch (commented):
This is a first-pass review, with some notes intended to help @necrodome join the conversation.

- SuperThis is somewhat equivalent to a context object, in Go context.Context is a reasonable analog
- Existing 'SuperThis' name is problematic, alternate under consideration "fpContext"
- As a coding convention, we would standarize on this as the first argument when wiring up the codebase
- Previously we passed a logger instance to most places, and now the logger is inside sthis, so many of the changes in this PR are of the form `s/logger/sthis/`
- I can report that this branch basically works (last tested by me before all commits were present)

## My open questions/concerns

- It would be great to remove the extraneous stuff from this PR, specifically fragmentation stuff is unrelated
- What is the long-term plan for RemoteMetaStore (lots of refs just commented out)
- Is there any cleaner/standard way of doing this is JS/TS world?  Are there are libraries that do something similar that we can point to?
- I would like to clarify the life-cycle of SuperThis.  In Go in particular, you are advised to not store context objects inside structs, and I believe the context objects are themselves immutable (modifying it creates a new instance).  This detail will become important if we are using it for things like instrumentation/tracing which may contain state (ie, current active span).  I ask this because I see mostly pass sthis during initialization and stash it inside the struct.  This may also relate to the clone() method, would like it's usage clarified.
- I interpretted jchris's guidance to favor moving quickly at the moment, so I also want to be cognizant of overdoing this review

mschoch:
Alright I thought maybe I could fix up some minor things and get it usable again, but seems to be more significant than I realized.

mschoch:
For the time being I'm basing future work on `eb8b739` as that was the last reliable SHA for me on this branch.  And I was expecting we'd get agreement on something close to that merged.  There's going to be some rework to pull all this together no matter which way we go.

src/runtime/gateways/file/utils.ts
```diff
@@ -1,46 +1,53 @@
...
   // .replace(/\?.*$/, "");
   const name = url.getParam("name");
   if (name) {
     const version = url.getParam("version");
-    if (!version) throw logger.Error().Url(url).Msg(`version not found`).AsError();
-    return SysContainer.join(basePath, version, name);
+    if (!version) throw sthis.logger.Error().Url(url).Msg(`version not found`).AsError();
+    return sthis.pathOps.join(basePath, version, name);
   }
-  return SysContainer.join(basePath);
```
jchris: this looks like the action

jchris:
for integration work with connect repo I am using 22d92213fd for now

jchris:
is this part of #185 now? I think we should merge both

tests/fireproof/crdt.test.ts
```diff
@@ -16,10 +16,13 @@ describe("Fresh crdt", function () {
     await sthis.start();
     const dbOpts: DatabaseOpts = {
       // meta: opts?.meta,
+      factoryUnreg() {
+        return Promise.resolve();
+      },
```
jchris: I'm not sure this is right

tests/fireproof/database.test.ts
```diff
@@ -51,7 +51,7 @@ describe("basic Database with record", function () {
   });
   beforeEach(async function () {
     await sthis.start();
-    db = Database.factory();
+    db = Database.factory("factory-name");
```
jchris: maybe this is temporary

src/blockstore/meta-key-helper.ts
```diff
@@ -1,14 +1,14 @@
...
-import type { DbMeta, CarClockHead, DbMetaEventBlock, CarClockLink } from "./types.js";
+import type { DbMeta, CarClockHead, DbMetaEventBlock, CarClockLink, DbMetaEvent } from "./types.js";
 import { SuperThis, CRDTEntry } from "../types.js";
 import { CID, Link } from "multiformats";
 import { base64pad } from "multiformats/bases/base64";
 import { Result, URI } from "@adviser/cement";
 import { getKeyBag } from "../runtime/key-bag.js";
 
-export async function decodeGatewayMetaBytesToDbMeta(sthis: SuperThis, byteHeads: Uint8Array) {
+export async function decodeGatewayMetaBytesToDbMeta(sthis: SuperThis, byteHeads: Uint8Array): Promise<DbMetaEvent[]> {
```
jchris, line 11: 👀

src/blockstore/store-remote.ts
```diff
@@ -31,12 +31,12 @@ import { SuperThis } from "../types.js";
...
 
 export async function RemoteDataStore(sthis: SuperThis, name: string, url: URI, opts: StoreOpts) {
-  const ds = new DataStoreImpl(sthis, name, url, opts);
+  const ds = new DataStoreImpl(sthis, url, opts);
   await ds.start();
   return ds;
 }
 export async function RemoteMetaStore(sthis: SuperThis, name: string, url: URI, opts: StoreOpts) {
-  const ms = new MetaStoreImpl(sthis, name, url, opts /* , true*/);
+  const ms = new MetaStoreImpl(sthis, url, opts);
```
jchris, line 39: will look for the other side of that `true` :)

src/crdt.ts
```diff
@@ -68,23 +69,27 @@ export class CRDT<T extends DocTypes> {
...
       meta: this.opts.meta,
-      threshold: this.opts.threshold,
+      // threshold: this.opts.threshold,
     });
     this.indexBlockstore = blockstoreFactory(sthis, {
-      name: name,
+      // name: opts.name,
       applyMeta: async (meta: TransactionMeta) => {
-        const idxCarMeta = meta as IdxMetaMap;
+        const idxCarMeta = meta as IndexTransactionMeta;
```
jchris, line 83: 👍

src/database.ts
```diff
@@ -1,92 +1,290 @@
...
+      });
+      db.onClosed(() => {
+        databases.unget(key);
+      });
+      return db;
+    }),
+  );
+}
+
+export class DatabaseShell<DT extends DocTypes = NonNullable<unknown>> implements Database<DT> {
```
jchris, line 131: I wanna understand what motivates this layer

src/runtime/gateways/file/node-filesystem.ts
```diff
@@ -17,7 +16,7 @@ export class NodeFileSystem implements SysFileSystem {
   };
 
   async start(): Promise<SysFileSystem> {
-    this.fs = await import("node:fs/promises");
+    this.fs = await import("fs/promises");
```
jchris, line 19: 👍 💯

mabels: lets use node: to ensure that deno will fail -;)

src/database.ts
mabels: close need to be refcounted 
if you open multiple times the same database like:
db1 = fireproof("test")
db2 = fireproof("test")
db3 = fireproof("test")
db4 = fireproof("test")
db1.close()
db2.close()
db4.close()
db3.close() // realclose is done. w
then close should do nothing until the last instance is closed.
We had some test which tested for that behaviour.

src/blockstore/store-remote.ts
mabels: if we need this flags it should be part of storeopts not explicit

src/runtime/gateways/memory/version.ts
```diff
@@ -0,0 +1 @@
+export const MEMORY_VERSION = "v0.19-memory";
```
jchris, line 1: is this supposed to be checked in?

tests/fireproof/fireproof.test.ts
```diff
@@ -84,7 +97,7 @@ describe("public API", function () {
   });
   it("should be a database instance", function () {
     expect(db).toBeTruthy();
-    expect(db instanceof Database).toBeTruthy();
+    expect(isDatabase(db)).toBeTruthy();
```
jchris, line 100: not sure our downstream users can be expected not to use `instanceof` maybe we should have both assertions? I'm not sure what isDatabase would be used for outside of this, so maybe we remove the test?

jchris (approved):
thanks!

PR #184, opened 2024-08-20
mprove the storekey url parameter handling

mabels:
There had been some oddities about when a key in the key-bag is created.

This should be now like:
if you pass not as storekey is created with the pattern
-  @[dbname](:idx):[data|meta|wal]@
if you pass a storekey with a starting and ending '@'
- the key is lookup and if not found it created from a random source
if you pass a btcbase58 z leading string than this is used as key!
if you pass insecure no encryption and decryption is used

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
   },
   "homepage": "https://github.com/fireproof-storage/fireproof#readme",
   "dependencies": {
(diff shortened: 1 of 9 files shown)


mabels:
Super-This includes this

PR #185, opened 2024-08-25
Add a Fragmationlayer into the Gateway Implementation

mabels:
Add a Fragmationlayer into the Gateway Implementation
           Changed dag-cbor -> cbor to safe some bytes [WARNING] file-format change

Usage is done via start from gateway just add "fragSize" parameter to URL like:
```
   url.defParam("fragSize", "1024")
```
The value sets the max blocksize which will passed than to the gateway.


mabels:
Sorry for making it base on SuperThis so do better:
git diff super-this

mschoch:
> Sorry for making it base on SuperThis

~~Can we retarget this PR to the `super-this` branch so that we can review the smaller diff here? (I expect we would merge SuperThis first anyway)~~

I see now that you pushed to that branch as well :(

Screenshot 2024-08-26 at 11 33 40 AM

jchris:
How is this related to the existing threshold? https://github.com/fireproof-storage/fireproof/blob/bdc2398adaa68b1c836d531b1282f39db3083010/src/blockstore/commitor.ts#L160

mschoch:
> How is this related to the existing threshold? 

I'll let @mabels answer directly, but yes @necrodome had pointed me to this configuration in Discord on Friday.  When @mabels and I discussed it, we identified a few pros/cons of this being done inside fireproof vs inside the gateway.

My short term motivation for looking into this was that the PartyKit storage is limited to values of around 128k.  This is so small it prevents our extremely simple fireproof tests from running correctly with the PartyKit adapter.

Having that threshold being configurable via the connection URL would be useful for our tests (since we already have a capability to run all tests with a custom ` FP_STORAGE_URL`.  But, from an end-user usability perspective that would be undesirable, you'd really want the implementation to communicate it's own limitations.  And we've also been looking towards a future with multiple remotes, and each one might have a different size limitation.  Solving that fully raises even more questions we're not prepared to answer.

If I can easily configure threshold for running the PartyKit gateway compliance tests and validate it works, I'd move on to the next thing.

jchris:
As I understand it now the fragmentor is at a lower layer than threshold. The threshold parameter splits into mulitple content addressed car files (with duplicate headers), the fragmentor cuts a given file in to slices sized X. It makes sense to support both as long as with such an approach we don't write too many features we dont need or add bundle bloat over time.

I got the partykit thing working with threshold by hardcoding it https://github.com/fireproof-storage/connect/pull/7#issuecomment-2322966330

So if we add config for threshold or just use this, they should both pass.

jchris (approved):
we should merge this PR and the super this stuff, in whatever order makes sense for merging.

PR #199, opened 2024-09-03
update the URI handling for party-kit

mabels:
Update cement for party-kit handling url like:
partykit://localhost:8993/


jchris (approved):
makes sense

PR #200, opened 2024-09-05
test that onload runs

jchris:
this should be ready to merge I just want to make sure the CI is clean

### DIFF
--- a/src/blockstore/types.ts
+++ b/src/blockstore/types.ts
@@ -248,6 +248,7 @@ export interface MetaStore extends BaseStore {
   load(branch?: string): Promise<DbMeta[] | Falsy>;
   // branch is defaulted to "main"
   save(meta: DbMeta, branch?: string): Promise<Result<void>>;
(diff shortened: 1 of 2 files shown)


src/blockstore/types.ts
```diff
@@ -248,6 +248,7 @@ export interface MetaStore extends BaseStore {
   load(branch?: string): Promise<DbMeta[] | Falsy>;
   // branch is defaulted to "main"
   save(meta: DbMeta, branch?: string): Promise<Result<void>>;
+  onLoad(branch: string, loadHandler: LoadHandler): () => void;
```
jchris, line 251: We use this interface in the connectors https://github.com/fireproof-storage/fireproof/blob/094d2a659a6ecaa271c4b0f8ef6659e2ba31141a/src/blockstore/connection-base.ts#L75

I want to test that we call `handleDbMetasFromStore` but I think that is not for this unit test.

PR #202, opened 2024-09-07
store parents in the header

jchris:
This changes the encoding so all the connectors can support multi-writer easily.

It basically brings what was in the wire encoding for netlify / aws / partykit etc into the core.

### DIFF
--- a/scripts/connect-partykit/connect-partykit.ts
+++ b/scripts/connect-partykit/connect-partykit.ts
@@ -46,6 +46,7 @@ export class ConnectPartyKit extends Connection {
         const base64String = event.data;
         const uint8ArrayBuffer = Base64.toUint8Array(base64String);
         const eventBlock = await this.decodeEventBlock(uint8ArrayBuffer);
+        // there should be a callback in the MetaStore that we can call from places like this
         await this.taskManager!.handleEvent(eventBlock);
         // @ts-ignore
         this.messageResolve?.([eventBlock.value.data.dbMeta as Uint8Array]);
--- a/src/blockstore/store.ts
+++ b/src/blockstore/store.ts
@@ -17,6 +17,9 @@ import type {
   LoadHandler,
   KeyedCrypto,
   Loadable,
+  CarClockHead,
+  DbMetaEventBlock,
+  CarClockLink,
 } from "./types.js";
 import { Falsy, StoreType, SuperThis, throwFalsy } from "../types.js";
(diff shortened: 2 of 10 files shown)


src/blockstore/store.ts
```diff
@@ -465,3 +508,48 @@ export class WALStoreImpl extends BaseStoreImpl implements WALStore {
     return this.gateway.destroy(this.url());
   }
 }
+
```
jchris, line 522: I had trouble with the node module in the smoke test

jchris: https://github.com/fireproof-storage/fireproof/actions/runs/10751859938/job/29819554791

tests/blockstore/store.test.ts
```diff
@@ -152,12 +154,19 @@ describe("MetaStore with a saved header", function () {
...
-    expect(header.cars[0]["/"]).toEqual(cid.toString());
+    expect(header).toBeDefined();
+    expect(header.parents).toBeDefined();
+
+    const decodedHeader = await (
+      store as unknown as { decodeMetaBlocks: (file: Uint8Array) => Promise<bs.DbMeta> }
+    ).decodeMetaBlocks(bytes);
+    expect(decodedHeader).toBeDefined();
+    expect(decodedHeader.cars).toBeDefined();
+    expect(decodedHeader.cars[0].toString()).toEqual(cid.toString());
```
jchris, line 166: how this is twice as long now... I bet we can flat it to one encoding

jchris: I want the cid and parents to be in json, maybe we should just bring the cars up to the json also -- json means we don't need imports in our server code

jchris:
Will make other connectors easier https://github.com/fireproof-storage/connect/pull/7#issuecomment-2336522649

scripts/connect-partykit/connect-partykit.ts
```diff
@@ -46,6 +46,7 @@ export class ConnectPartyKit extends Connection {
         const base64String = event.data;
         const uint8ArrayBuffer = Base64.toUint8Array(base64String);
         const eventBlock = await this.decodeEventBlock(uint8ArrayBuffer);
+        // there should be a callback in the MetaStore that we can call from places like this
```
jchris, line 49: @valorant-dhruv you'll need this when it comes time to listen to the websocket

jchris:
I think this has what is needed to connect the meta with parent cid tracking

PR #210, opened 2024-09-11
subscribe

jchris:
to support gateway events

### DIFF
--- a/src/blockstore/store.ts
+++ b/src/blockstore/store.ts
@@ -152,8 +152,7 @@ export class MetaStoreImpl extends BaseStoreImpl implements MetaStore {
       ensureLogger(sthis, "MetaStoreImpl"),
     );
     if (remote && opts.gateway.subscribe) {
-      this.logger.Debug().Str("gateway", opts.gateway.subscribe?.toString()).Msg("Gateway information");
-      this.logger.Debug().Str("url", url.toString()).Msg("Subscribing to the gateway with URL");
+      this.logger.Debug().Str("url", url.toString()).Msg("Subscribing to the gateway");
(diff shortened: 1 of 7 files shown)


src/blockstore/store.ts
```diff
@@ -151,6 +151,11 @@ export class MetaStoreImpl extends BaseStoreImpl implements MetaStore {
       sthis,
       ensureLogger(sthis, "MetaStoreImpl"),
     );
+    this.logger.Debug().Str("gateway", opts.gateway.subscribe?.toString()).Msg("Gateway information");
+    if (remote && opts.gateway.subscribe) {
+      this.logger.Debug().Str("url", url.toString()).Msg("Subscribing to the gateway with URL");
+      opts.gateway.subscribe(url, (byteHead: Uint8Array) => this.handleEventByteHead(byteHead));
+    }
```
jchris, line 157: this subscribe is only called on remote blockstores... this could be implemented as a subclass thing on the meta or something, but I think the way I have it is good enough for testing the runtime.

mschoch (approved):
Looks good to me, two minor questions.

Should gateways expect to only have one subscriber, or should they support multiple invocations of this method.  If we only expect 1 for the life of the gateway, it might be cleaner if it was passed in with start()

Did this change break the test you disabled?

src/blockstore/store.ts
jchris: this is a single event, not the multi-event thing that comes from the GET. that's not essential, we could wrap this as a multi if easier

jchris:
i've modified the pk gateway to connect on subscribe, not start, so this only connects when used as a remote store

PR #212, opened 2024-09-11
move to multi-meta

jchris:


jchris:
here is a quick and dirty take at #211 for just getting it working. I will try to bring connectors to functional parity using this, and then upgrading to #211 should be a matter of encoding if necessary.

mschoch (approved):

src/types.ts
```diff
@@ -262,3 +262,9 @@ export interface DocResponse {
 export type UpdateListenerFn<T extends DocTypes> = (docs: DocWithId<T>[]) => Promise<void> | void;
 export type NoUpdateListenerFn = () => Promise<void> | void;
 export type ListenerFn<T extends DocTypes> = UpdateListenerFn<T> | NoUpdateListenerFn;
+
+export interface CRDTEntry {
+  data: string;
```
mabels, line 267: readonly!

PR #213, opened 2024-09-12
Mmain

jchris:
improvements (passing partykit tests)

also merges subscribe branch. 

all this can be merged to main after review


mschoch (approved):
I'm fine with all these changes, but note that we disabled one additional test, and I would prefer to have CI passing before we merge.

src/blockstore/store-factory.ts
```diff
@@ -141,24 +141,24 @@ export function registerStoreProtocol(item: GatewayFactoryItem): () => void {
 //   })
 // }
 
-// const onceDataStoreFactory = new KeyedResolvOnce<DataStoreImpl>();
+const onceDataStoreFactory = new KeyedResolvOnce<DataStoreImpl>();
```
jchris, line 144: this passes tests both ways but logs a lot less noise with it in

jchris:
merging via the other PRS first

PR #226, opened 2024-09-18
FP_FORMAT for better log formatting

mabels:
by setting of FP_FORMAT env to:

- jsonice makes the log output in multiline json
- yaml makes the log output in yaml
- json makes the log output in singleline json (default)


jchris:
thanks!

PR #229, opened 2024-09-19
update Connectable to include sthis

mschoch:
and export it from fireproof

### DIFF
--- a/src/blockstore/connection-base.ts
+++ b/src/blockstore/connection-base.ts
@@ -1,6 +1,6 @@
 import { Logger, URI } from "@adviser/cement";
 
-import { throwFalsy } from "../types.js";
+import {type SuperThis, throwFalsy} from "../types.js";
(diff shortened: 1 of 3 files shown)


mschoch:
@jchris this was what I tried to make Connectable usable again in the connect repo.

jchris:
thanks I will preview it when i get into that layer which is soon

mschoch:
I realize now this is wrong.  We export the blockstore package under the name 'bs', but I didn't realize that until now.

src/blockstore/connection-base.ts
```diff
@@ -14,6 +14,7 @@ export interface Connectable {
     readonly ebOpts: BlockstoreRuntime;
   };
   readonly name?: string;
+  readonly sthis: SuperThis;
```
jchris, line 17: what about this?

PR #230, opened 2024-09-19
make keys extractable

mabels:
look into the test in tests/blockstore/keyed-crypto.test.ts test "extract keyMaterial" howto to use it.


jchris (approved):
this looks helpful thanks

PR #231, opened 2024-09-19
subscribe test

jchris:


src/blockstore/fragment-gateway.ts
```diff
@@ -171,7 +171,27 @@ export class FragmentGateway implements bs.Gateway {
     if (this.innerGW.subscribe) {
       return this.innerGW.subscribe(url, callback);
     } else {
-      return Result.Err(this.logger.Error().Msg("Subscribe not supported").AsError());
+      let lastData: Uint8Array | undefined = undefined;
+      let interval = 100;
+      const fetchData = async () => {
+        const result = await this.innerGW.get(url);
```
jchris, line 177: @mabels this code is definitely in the wrong place, but it works just right -- fallback pattern for gateways that dont have a push capabiliity on the server or haven't implemented it yet

PR #233, opened 2024-09-20
enable rough support of deno

mabels:
This is not complete but with an odd deno.json deno is running.


jchris:
what is odd on the deno.json? does this make senes to put in CI once it is green?

mabels:
the odd is i generate a deno.json in our npm package:

deno run --config node_modules/@fireproof/core/deno.json --allow-read --allow-write --allow-env --unstable-sloppy-imports ./node-test.ts

PR #234, opened 2024-09-20
export decode/encode from meta store

jchris:
this supports gateway re-encoding


src/blockstore/types.ts
```diff
@@ -196,7 +196,7 @@ export interface CommitOpts {
 
 export interface DbMeta {
   readonly cars: CarGroup;
-  // key?: string;
+  key?: string;
```
jchris, line 199: @mabels I'm ok to do some type overlay here if you want to adjust it.

PR #278, opened 2024-10-07
cleanup for super-this merge

mabels:


mabels:
I changed the return value of gateway subscription to enable a unregister.

PR #280, opened 2024-10-07
Store-correctness

jchris:
this should help avoid pushing meta until all the car files have been uploaded. Things worth considering here:
- retry timing and limit
- compact before send, all the operations and noLoaderOps car files into one car file, then send with the latest dbMeta


jchris:
@mabels you can merge this anytime, but I'd love help with the other things listed

jchris:
doing compaction will be more work, it will require:
- at beginning of `_doProcess` coalece all operations and noLoaderOps into a single car file (or files if fragment or threshold are active). file uploads should not be modified.
- upload the new car file, with the latest db meta pointing to it. the header of the new car file should be from the most recent car file in the coalced group.
- coalce can be block level (default compaction) not database compaction

PR #283, opened 2024-10-08
compact before sync push

jchris:
this runs weird but might help someone


jchris:
solved by https://github.com/fireproof-storage/fireproof/pull/286

PR #286, opened 2024-10-13
only commit clock transactions on local update

jchris:


jchris:
this is worth backporting to 0.19

PR #287, opened 2024-10-13
backport group clock optimization

jchris:


src/blockstore/transaction.ts
```diff
@@ -137,6 +141,27 @@ export class BaseBlockstore implements BlockFetcher {
...
+
+  async commitTransaction<M extends TransactionMeta>(
+    t: CarTransaction,
+    done: M,
+    opts: CarTransactionOpts,
+  ): Promise<TransactionWrapper<M>> {
+    if (!this.loader) throw this.logger.Error().Msg("loader required to commit").AsError();
+    const cars = await this.loader?.commit<M>(t, done, opts);
+    if (this.ebOpts.autoCompact && this.loader.carLog.length > this.ebOpts.autoCompact) {
+      setTimeout(() => void this.compact(), 10);
```
jchris, line 156: this timeout is copied from the commit() function below. It's only every 100 writes, so this is just to delay after the current write finishes

PR #299, opened 2024-10-15
initial public api spec

jchris:
This is the test we expect all versions of 1.x to pass. If we break this test, we are talking about 2.0

Let's review and discuss before we merge. These are the main feature areas to specify:

- Query options and response object shape for `db.query()`, `db.allDocs()`, `db.changes()` and `db.subscribe()` can be normalized.
- Metadata like `doc._id` are implicitly specified here, we can do better
- We should specify `_proof` or something similar to give access to the underlying Merkle integrity
- The history API will come post 1.0 but this is a good context to think about it. 

Adding assertions (not removing them) from this test is how we know we are getting close to a new 1.1, 1.2, etc

### DIFF
--- a/tests/fireproof/feature-spec.test.ts
+++ b/tests/fireproof/feature-spec.test.ts
@@ -1,0 +1,135 @@
+import { mockSuperThis } from "../helpers.js";
+import { CID } from "multiformats/cid";
+
+import { Database, DatabaseFactory, DocResponse, DocWithId, IndexRows, bs, fireproof, index, isDatabase } from "@fireproof/core";
+
+export function carLogIncludesGroup(list: bs.AnyLink[], cid: CID) {
+  return list.some((c) => c.equals(cid));
+}
+
+interface FooType {
+  readonly foo: string;
+}
+
+describe("public API", function () {
+  interface Doc {
+    foo: string;
+  }
+  let db: Database;
+  let ok: DocResponse;
+  let doc: DocWithId<Doc>;
+  let query: IndexRows<string, Doc>;
+  const sthis = mockSuperThis();
+
+  afterEach(async () => {
+    await db.close();
+    await db.destroy();
+  });
+
+  beforeEach(async function () {
+    await sthis.start();
+    db = fireproof("test-api");
+    // index = index(db, 'test-index', (doc) => doc.foo)
+    ok = await db.put({ _id: "test", foo: "bar" });
+    doc = await db.get("test");
+    query = await db.query<string, Doc>((doc) => doc.foo);
+  });
+  it("should be a database instance", function () {
+    expect(db).toBeTruthy();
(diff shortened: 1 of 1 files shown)


jchris:
Also relevant are questions about renaming "database" to "document" in which case what do we call the return of `const val = document.get('id')`?

Candidate names are `val` (value) `rec` (record) `it` (item), any others?

tests/fireproof/feature-spec.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  let doc: DocWithId<Doc>;
+  let query: IndexRows<string, Doc>;
+  const sthis = mockSuperThis();
+
+  afterEach(async () => {
+    await db.close();
+    await db.destroy();
+  });
+
+  beforeEach(async function () {
```
mabels, line 29: async () =>

tests/fireproof/feature-spec.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    await sthis.start();
+    db = fireproof("test-api");
+    // index = index(db, 'test-index', (doc) => doc.foo)
+    ok = await db.put({ _id: "test", foo: "bar" });
+    doc = await db.get("test");
+    query = await db.query<string, Doc>((doc) => doc.foo);
+  });
+  it("should be a database instance", function () {
```
mabels, line 37: function  =>

tests/fireproof/feature-spec.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  const sthis = mockSuperThis();
+  afterEach(async function () {
+    await db.close().catch(() => true);
+    await db.destroy();
+  });
+  beforeEach(async function () {
+    // erase the existing test data
+    await sthis.start();
+
+    db = DatabaseFactory("test-reopen-spec", { autoCompact: 100000 });
```
mabels, line 117: why not fireproof?

mabels (commented):
Nice

mabels:
i would go for this:
```
const tdb = db.typed<DocType>() // this will not break with the old api
interface TypedItem<T> {
    meta: Meta
    data: T
}
interface TypedDatabase<T> {
    get(key): Result<TypedItem>
    put(typedItem: T| TypedItem<T>): Result<TypedItem<T>>
    query(key: T|AllDocs): Result<TypedItem<T>[]> // we don't need AllDocs
    subscribe(key: T|AllDocs, CB: (TypedItem<T>[]) => Result<void>): Result<Registration>
    // do we need changes?
}
```

jchris:
I like the suggestion of `collection` or `ledger` for what we call "database" now

mabels:
Please add the iterator and readstream api to alldocs/query.
If you need help --- ask

PR #315, opened 2024-10-24
Query-api-cleanup

jchris:
Normalize `query` `allDocs` and `changes` responses


jchris:
we should make includeDocs true by default

mabels:
substituted by:
https://github.com/fireproof-storage/fireproof/pull/464

PR #341, opened 2024-11-07
change useDocument to take an object instead of a function

necrodome:
I don't think the optimisation made here is worth the worse DX. We can also provide both options but most will try the object

PR for discussions


mabels:
Thx for this @necrodome. I will also be available in >=0.19.112.

PR #348, opened 2024-11-11
unified wireprotocol

mabels:
pls review and share any ideas


src/protocols/fp-car.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+/*
+ * The response to a GetCar request.
+ * As JSON, this is:
+ * {
+ *    "type": "get-car-res",
+ *    "tid": "123",
+ *    "payload": {
+ *       "cid": "bafybeib2v"
+ *       "bytes": [0x65, 0x66, 0x67, ...],
```
jchris, line 38: why not just send the car raw as binary? the CAR get and PUT dont need to be parsed by the server

src/protocols/fp-car.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+/*
+ * A request to get a CAR file.
+ * As JSON, this is:
+ * {
+ *    "type": "put-car-req",
+ *    "tid": "123",
+ *    "payload": {
+ *       "cid": "bafybeib2v"
+ *       "bytes": [0x65, 0x66, 0x67, ...],
```
jchris, line 61: the put request to the signed upload URL generator doesnt include the bytes, the bytes are sent raw on the s3 protocol on a subsequent request

src/protocols/fp-car.ts
```diff
@@ -0,0 +1,137 @@
...
+export interface FPEnvelopePutCarReq extends FPCarReq<FPPutCarReq> {
+    readonly type: FPEnvelopeType.PUT_CARREQ;
+    readonly storeType: FPStoreType.CAR;
+}
+
+export interface FPDelCarReq {
+    readonly cid: CID;
+}
+/*
+ * The response to a PutCar request.
```
jchris, line 76: the signed upload url can be returned as a redirect, but today we put it in a json envelope, either is ok, but the result here should just be a URL in the s3 protocol

src/protocols/fp-file.ts
```diff
@@ -0,0 +1,144 @@
...
+ * {
+ *    "type": "get-file-req",
+ *    "tid": "123",
+ *    "payload": {
+ *       "cid": "bafybeib2v"
+ *    }
+ *    "auth": { ... }
+ *    "meta": { ... }
+ * }
+ */
```
jchris, line 20: we need to be able to serve unencrpyted files directly at eg: `` with correct content type

jchris: being able to PUT raw files is not required

src/protocols/fp-meta.ts
```diff
@@ -0,0 +1,137 @@
+import { FPEnvelopeTurnaround, FPEnvelopeType, FPStoreType } from "./fp-envelope";
+import { CID } from "multiformats";
+
+export interface FPGetMetaReq {
+    readonly cid: CID;
+}
+/*
+ * A request to get a META file.
```
jchris, line 8: meta isn't quite a file, its an API response, as it must be composed by the server dynamically depending on current state

src/protocols/fp-meta.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+/*
+ * The response to a GetMeta request.
+ * As JSON, this is:
+ * {
+ *    "type": "get-meta-res",
+ *    "tid": "123",
+ *    "payload": {
+ *       "cid": "bafybeib2v"
+ *       "bytes": [0x65, 0x66, 0x67, ...],
```
jchris, line 37: this might be worth taking down another layer as JSON, instead of embedding bytes. Ideally you can see the CIDs of the relevant CARs in the JSON here, it would not be a larger serialization.

src/protocols/fp-meta.ts
```diff
@@ -0,0 +1,137 @@
...
+    readonly cid: CID;
+    readonly bytes: Uint8Array;
+}
+/*
+ * A request to get a META file.
+ * As JSON, this is:
+ * {
+ *    "type": "put-meta-req",
+ *    "tid": "123",
+ *    "payload": {
```
jchris, line 58: this doesn't have a body i dont think

src/protocols/fp-meta.ts
```diff
@@ -0,0 +1,137 @@
...
+export interface FPEnvelopePutMetaReq extends FPMetaReq<FPPutMetaReq> {
+    readonly type: FPEnvelopeType.PUT_METAREQ;
+    readonly storeType: FPStoreType.META;
+}
+
+export interface FPDelMetaReq {
+    readonly cid: CID;
+}
+/*
+ * The response to a PutMeta request.
```
jchris, line 75: this should respond with the HEAD post-PUT, inclusive of the PUTs effect -- same as a GET.

src/protocols/fp-wal.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export type FPGetWALReq = WALState
+/*
+ * A request to get a WAL file.
+ * As JSON, this is:
+ * {
+ *    "type": "get-wal-req",
+ *    "tid": "123",
+ *    "payload": {
+ *   operations: DbMeta[];
```
jchris, line 13: right now this means we need a WAL per remote. we could keep high water marks, but that would require a function from (encrypted) CRDT log events to carLog CIDs. Not hard to do ...  but wait til after we have the next layer of testing.

jchris: this might be a good place for @damienkatz help

PR #350, opened 2024-11-12
Relax indexeddb name sanitization to include hypens

necrodome:
Probably can include more characters but for now it would be nice to have at least hypen for a valid char since uuids are used for remote names by default.

on dashboard, indexeddb is used as source of truth for the database list

### DIFF
--- a/src/runtime/gateways/indexdb/gateway.ts
+++ b/src/runtime/gateways/indexdb/gateway.ts
@@ -63,7 +63,7 @@ export interface DbName {
 
 function joinDBName(...names: string[]): string {
   return names
-    .map((i) => i.replace(/^[^a-zA-Z0-9]+/g, "").replace(/[^a-zA-Z0-9]+/g, "_"))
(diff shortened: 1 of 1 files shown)


src/runtime/gateways/indexdb/gateway.ts
```diff
@@ -63,7 +63,7 @@ export interface DbName {
 
 function joinDBName(...names: string[]): string {
   return names
-    .map((i) => i.replace(/^[^a-zA-Z0-9]+/g, "").replace(/[^a-zA-Z0-9]+/g, "_"))
+    .map((i) => i.replace(/^[^a-zA-Z0-9]+/g, "").replace(/[^a-zA-Z0-9-]+/g, "_"))
```
jchris, line 66: maybe worth testing cuz in the test suite

mabels:
That's ok without test ---
All extended thought are not so imported now.
Like having ledger names with Emojis or in Kanji.
If we want to address that a transformation of the name to base58btc might be the right solution. And we need a generic ledger iteration method.

PR #352, opened 2024-11-14
genesis block --- or mergable by default

jchris:
This is a test that two differently originated databases can merge.

### DIFF
--- a/tests/fireproof/fireproof.test.ts
+++ b/tests/fireproof/fireproof.test.ts
@@ -45,7 +45,7 @@ describe("dreamcode", function () {
   });
   beforeEach(async function () {
     await sthis.start();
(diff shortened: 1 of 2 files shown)


tests/fireproof/fireproof.test.ts
```diff
@@ -45,7 +46,7 @@ describe("dreamcode", function () {
   });
   beforeEach(async function () {
     await sthis.start();
-    db = fireproof("test-db");
+    db = fireproof("test-db", { public: true });
```
jchris, line 49: this is not the correct way to configure `public:true` but we should probably make this work instead of requiring the keyname trick

mabels:
In series_0.20  the url's are preprocessed there we could map public into
key=insecure.

PR #353, opened 2024-11-15
use-fireproof gets a forwarding package

mabels:
This is needed due to the runtime problems caused by doubling the code if you
use use-fireproof and fireproof/core at the same time.


mabels:
there is now a v0.19.112-dev-usefp which hopefully solve the problem

PR #467, opened 2024-12-19
write multiple documents in a single operation

jchris:
this allows inserting multiple records into a single CRDT operation. For some workloads it may see a benefit, as larger datasets can have fewer log entries.

## Summary by CodeRabbit

- **New Features**
  - Introduced a `bulk` method for processing multiple documents in the ledger.
  - Enhanced `WriteQueue` with bulk processing capabilities and improved error handling.
  - Added new configuration options for the write queue.

- **Bug Fixes**
  - Improved resource management in the `shellClose` method.

- **Tests**
  - Added a test case for the new `bulk` method in the ledger.
  - Restructured and renamed test cases for clarity in the `Ledger` functionality tests.
  - Standardized write queue configuration across multiple `CRDT` test cases.
  - Updated test configurations in the indexer test suite to utilize default write queue options.

### DIFF
--- a/src/write-queue.ts
+++ b/src/write-queue.ts
@@ -4,10 +4,12 @@ type WorkerFunction<T extends DocTypes> = (tasks: DocUpdate<T>[]) => Promise<Met
 
 export interface WriteQueue<T extends DocTypes> {
   push(task: DocUpdate<T>): Promise<MetaType>;
+  bulk(tasks: DocUpdate<T>[]): Promise<MetaType>;
 }
 
 interface WriteQueueItem<T extends DocTypes> {
-  readonly task: DocUpdate<T>;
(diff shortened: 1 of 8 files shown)


src/write-queue.ts
```diff
@@ -21,7 +23,7 @@ export function writeQueue<T extends DocTypes>(worker: WorkerFunction<T>, payloa
     isProcessing = true;
 
     const tasksToProcess = queue.splice(0, payload);
-    const updates = tasksToProcess.map((item) => item.task);
+    const updates = tasksToProcess.flatMap((item) => item.task || item.tasks || []);
```
jchris: this is ugly, why not just have task be `DocUpdate<T> | DocUpdate<T>[]`?

PR #544, opened 2025-01-17
rename ledger to database

jchris:
I am wary of making changes to our public API since we have a fair number of users a `ledger` release would break (and almost no one using the ledger naming) so this needs a little more review, but it basically undoes the ledger technical renaming. (This doesn't change how I feel about the concept of the ledger, and the docs will still talk about ledger but it's just not one of the APIs we make public in the code yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@ What are the breaking changes in the 0.20 series?
   from passing Uint8Arrays and Url to passing a runtime Object which
   discribes the data on a semantic level. This give the gateway the
   ability to serialize and deserialize the data and enhance if needed.
-- We rename from Database to Ledger with no functional changes.
+- We rename from Database to Database with no functional changes.
 - a new memory implementation is replacing the memfs implementation.
   The prev use was to add to file:// the parameter fs=memfs now just use
-  memory:// works for Ledger as KeyBags.
+  memory:// works for Database as KeyBags.
 - The opts parameter of fireproof changed so that it's now possible
   to specify a every persitence like data/file/meta/wal or index-data/index-file/index-meta/index-wal a own URL. In the internals we also
   prevents the multiple instanciation of Gateways and Stores with results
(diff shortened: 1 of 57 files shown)


CHANGELOG.md
```diff
@@ -11,10 +11,10 @@ What are the breaking changes in the 0.20 series?
   from passing Uint8Arrays and Url to passing a runtime Object which
   discribes the data on a semantic level. This give the gateway the
   ability to serialize and deserialize the data and enhance if needed.
-- We rename from Database to Ledger with no functional changes.
+- We rename from Database to Database with no functional changes.
```
jchris, line 14: Just remove this line altogether since we did not release?

package-use-fireproof.json
```diff
@@ -26,7 +26,7 @@
   "scripts": {},
   "author": "J Chris Anderson",
   "license": "Apache-2.0 OR MIT",
-  "gptdoc": "Fireproof/React/Usage: import { useLiveQuery, useDocument } from 'use-fireproof'; function App() { const result = useLiveQuery(doc => doc.word, { limit: 10 }); const [{ count }, setDoc, saveDoc] = useDocument({_id: 'count', count: 0}); return (<><p>{count} changes</p><input type='text' onChange={() => saveDoc({count: count + 1})} onSubmit={e => useLiveQuery.ledger.put({word: e.target.value})} /><ul>{result.map(row => (<li key={row.id}>{row.key}</li>))}</ul></>)}",
+  "gptdoc": "Fireproof/React/Usage: import { useLiveQuery, useDocument } from 'use-fireproof'; function App() { const result = useLiveQuery(doc => doc.word, { limit: 10 }); const [{ count }, setDoc, saveDoc] = useDocument({_id: 'count', count: 0}); return (<><p>{count} changes</p><input type='text' onChange={() => saveDoc({count: count + 1})} onSubmit={e => useLiveQuery.database.put({word: e.target.value})} /><ul>{result.map(row => (<li key={row.id}>{row.key}</li>))}</ul></>)}",
```
jchris, line 29: this is the main public API I don't want to break

jchris:
this reverts https://github.com/fireproof-storage/fireproof/pull/316 which was never released, sorry for the noise

PR #564, opened 2025-01-30
Streaming allDocs/query API

icidasset:
Adds the ability to stream or async iterate documents.

## New API

Old API is still there; In addition to that, this PR adds the new method:

`database.select()`
`database.select("field")`
`database.select(queryOpts)`
`database.select("field", queryOpts)`

### All docs behaviour with `.select()`

```ts
const db = fireproof("name")
const allDocs = db.select()

// AsyncGenerator that provides a snapshot of all documents
for await (const doc of allDocs.snapshot()) { ... }

// Alternatively, get all at once:
const docs = await allDocs.toArray()

// ReadableStream that provides a snapshot of all docs and listen for new docs
for await (const doc of allDocs.live()) { ... }

// Since ...
allDocs.snapshot({ since: db.clock })
allDocs.live({ since: db.clock })
await allDocs.toArray({ since: db.clock })

// ReadableStream that listens for new docs
for await (const doc of allDocs.future()) { ... }

// Subscribe with callback
const unsub = allDocs.subscribe(doc => ...)
```

### Query behaviour with `.select("field")`

```ts
const db = fireproof("name")
const query = db.select("field", { limit: 5 })

// AsyncGenerator that provides a snapshot of the docs matching the query
for await (const doc of query.snapshot()) { ... }

// ReadableStream that provides a snapshot of the query docs and listens for new docs matching the query
for await (const doc of query.live()) { ... }

// ReadableStream that listens for new docs matching the query
for await (const doc of query.future()) { ... }

// Subscribe with callback
const unsub = allDocs.subscribe(doc => ...)
```

## Other changes

- `ledger.clock` and `database.clock` return the current `crdt.clock.head`
- Unsub functionality for crdt clock.
- `Ledger.changes` and `Ledger.subscribe` have been removed.
- Had to translate various functions into generators.

## To do

- [x] Finalize new API
- [x] `excludeDocs` option
- [x] Include `key` and `value` from prolly tree in all results
- [x] Add back old API and move new API to `.select()` method (which merges `allDocs` and `query`)
- [ ] Make sure new `allDocs` flow in `.select()` uses query options
- [x] Make sure new `query` flow in `.select()` uses query options
- [x] Make sure all tests pass
- [ ] Add `row.clock` property?

## Unclear

It's not clear what `value` should be in the "allDocs" flow (ie. when not specifying a field and thus listing all documents). I made this a list of all "values" in the document. For example, given the doc `{ a: "A", b: "B" }`, the value will be `[ "A", "B" ]`

## Summary by CodeRabbit

- **New Features**
  - Enhanced real-time updates and streaming capabilities, offering more flexible query options and live data synchronization for an improved experience.
  - Introduced new asynchronous utility for handling iterable arrays.
  
- **Refactor**
  - Optimized document retrieval, indexing, and change processing routines to boost performance and reliability.
  - Updated type definitions for improved flexibility in handling documents and queries.

- **Tests**
  - Expanded test coverage for streaming and querying functionalities to ensure robust, consistent behavior.
  - Added comprehensive tests for the Streaming API to validate various functionalities.

### DIFF
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -413,7 +413,6 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
   async _updateIndex(): Promise<IndexTransactionMeta> {
     await this.ready();
     this.logger.Debug().Msg("enter _updateIndex");
-    console.log("📣 UPDATE INDEX");
     if (this.initError) throw this.initError;
     if (!this.mapFn) throw this.logger.Error().Msg("No map function defined").AsError();
     let rows: DocumentRow<K, T, R>[];
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -413,6 +413,7 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
   async _updateIndex(): Promise<IndexTransactionMeta> {
     await this.ready();
     this.logger.Debug().Msg("enter _updateIndex");
+    console.log("📣 UPDATE INDEX");
     if (this.initError) throw this.initError;
     if (!this.mapFn) throw this.logger.Error().Msg("No map function defined").AsError();
     let rows: DocumentRow<K, T, R>[];
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -200,7 +200,15 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
 
     return {
       snapshot: (sinceOpts) => this.#snapshot(qryOpts, sinceOpts, { waitFor }),
-      subscribe: (callback) => this.#subscribe(qryOpts, callback),
+      subscribe: (callback) =>
+        this.#subscribe(qryOpts, async (arg) => {
+          callback(arg);
+
+          // NOTE: Sometimes this chain of operations freezes when run in parallel,
+          //       hence the reason why we not execute the callback afterwards.
+          await this._updateIndex();
+          await this._hydrateIndex();
+        }),
       toArray: (sinceOpts) => arrayFromAsyncIterable(this.#snapshot(qryOpts, sinceOpts, { waitFor })),
 
       live(opts?: { since?: ClockHead }) {
@@ -306,10 +314,7 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
     callback: ((row: Row<K, R>) => void) | ((row: DocumentRow<K, T, R>) => void),
   ): () => void {
     // NOTE: Despite using onTick or onTock, it always loads the document (update).
-    const unsubscribe = this.crdt.clock.onTick(async (updates: DocUpdate<NonNullable<unknown>>[]) => {
-      await this._updateIndex();
-      await this._hydrateIndex();
-
+    const unsubscribe = this.crdt.clock.onTick((updates: DocUpdate<NonNullable<unknown>>[]) => {
       const mapFn = this.mapFn?.bind(this);
       if (!mapFn) throw this.logger.Error().Msg("No map function defined").AsError();
 
@@ -378,6 +383,11 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
           if (isClosed) return;
           controller.enqueue({ row, marker: { kind: "new" } });
         });
+      },
+
+      async pull() {
+        await updateIndex();
+        await hydrateIndex();
       },
 
       cancel() {
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -8,7 +8,6 @@ import {
   type DocFragment,
   type IdxMetaMap,
   type IndexKeyType,
-  type IndexRows,
   type DocTypes,
   type IndexUpdateString,
   throwFalsy,
@@ -20,6 +19,12 @@ import {
   type HasLogger,
   type HasSuperThis,
   type RefLedger,
+  QueryResponse,
+  ChangesOptions,
+  IndexRow,
+  InquiryResponse,
+  DocumentRow,
+  Row,
 } from "./types.js";
 // import { BaseBlockstore } from "./blockstore/index.js";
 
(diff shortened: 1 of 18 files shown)


icidasset:
@mabels This is the current state of what I have. Any thoughts on the API? See PR description.

src/indexer.ts
```diff
@@ -175,57 +179,139 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
...
-      return await applyQuery<K, T, R>(this.crdt, await throwFalsy(this.byKey.root).get(encodedKey), opts);
-    }
-    if (Array.isArray(opts.keys)) {
-      const results = await Promise.all(
-        opts.keys.map(async (key: DocFragment) => {
-          const encodedKey = encodeKey(key);
-          return (await applyQuery<K, T, R>(this.crdt, await throwFalsy(this.byKey.root).get(encodedKey), opts)).rows;
-        }),
+  query(opts: QueryOpts<K> = {}, { waitFor }: { waitFor?: Promise<unknown> } = {}): QueryResponse<T> {
+    const query = async (since?: ClockHead, sinceOptions?: ChangesOptions) => {
```
mabels: function

src/indexer.ts
```diff
@@ -175,57 +179,139 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
...
-      {
-        // @ts-expect-error getAllEntries returns a different type than range
-        result: all.result.map(({ key: [k, id], value }) => ({
-          key: k,
-          id,
-          value,
-        })),
+    };
+
+    const snapshot = (opts: { since?: ClockHead } & ChangesOptions = {}) => {
```
mabels: function

src/indexer.ts
```diff
@@ -175,57 +179,139 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
...
+
+        for await (const doc of await query(opts.since, opts)) {
+          if (doc) yield doc;
+        }
+      }
+
+      return docsWithId();
+    };
+
+    const subscribe = (callback: (doc: DocWithId<T>) => void) => {
```
mabels: function

src/indexer.ts
```diff
@@ -175,57 +179,139 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
...
+      const unsubscribe = this.crdt.clock.onTick((updates: DocUpdate<NonNullable<unknown>>[]) => {
```
mabels: function

src/indexer.ts
```diff
@@ -175,57 +179,139 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
...
+        updates.forEach((update) => {
+          callback(docUpdateToDocWithId(update as DocUpdate<T>));
+        });
+      });
+
+      return unsubscribe;
+    };
+
+    const stream = (opts: { futureOnly: boolean; since?: ClockHead } & ChangesOptions) => {
```
mabels: function

src/indexer.ts
```diff
@@ -175,57 +179,139 @@ export class Index<K extends IndexKeyType, T extends DocTypes, R extends DocFrag
...
-      const encodedKey = encodeKey(opts.key);
-      return await applyQuery<K, T, R>(this.crdt, await throwFalsy(this.byKey.root).get(encodedKey), opts);
-    }
-    if (Array.isArray(opts.keys)) {
-      const results = await Promise.all(
-        opts.keys.map(async (key: DocFragment) => {
-          const encodedKey = encodeKey(key);
-          return (await applyQuery<K, T, R>(this.crdt, await throwFalsy(this.byKey.root).get(encodedKey), opts)).rows;
-        }),
+  query(opts: QueryOpts<K> = {}, { waitFor }: { waitFor?: Promise<unknown> } = {}): QueryResponse<T> {
```
mabels: this method should splitted in a shorter one.

src/types.ts
```diff
@@ -257,26 +258,20 @@ export interface IdxMetaMap {
 export interface QueryOpts<K extends IndexKeyType> {
   readonly descending?: boolean;
   readonly limit?: number;
-  includeDocs?: boolean;
```
mabels: why you remove the includeDocs option?

icidasset: From #315 I've understood that `includeDocs` should be on by default. I can add the option back to exclude docs? That said, I don't understand the use-case, so some context here would be great.

https://github.com/fireproof-storage/fireproof/pull/315#issuecomment-2521246582

src/indexer.ts
icidasset: I'm depending on the `this` binding in many of these functions, so I prefer this notation to keep it short.

icidasset: Actually, let me rewrite this and create private methods on the parent class.

src/types.ts
icidasset: Talked about this with Chris & Meno (separately) in Discord. Looks like we want the `key` and `value` in the result as well (cc @mabels), so that'll be an adjustment to `allDocs`. 

Regarding `includeDocs`, ideally this would be an option to pass in when wanting to exclude the docs. Hopefully we can do some function overloading here so the types are nicer to work with.

src/indexer.ts
icidasset: @mabels See 032906d. That better?

src/indexer.ts
icidasset: Also addressed in 032906d

src/indexer.ts
mabels: Why better.

src/types.ts
mabels: yes --- rename it to excludeDocs so that the default is including the docs.

src/indexer.ts
icidasset: Shorter functions, less nested scoping and all that.

PR #575, opened 2025-02-07
new useDocument API

jchris:
maybe this can merge to main instead, but I can only test with the 0.20 preview line due to build issues


jchris:
example https://codepen.io/J-Chris-Anderson/pen/MYgNYdx?editors=0110

PR #599, opened 2025-02-21
error handling for undefined

jchris:
this is to fix the dev wart 

{"module":"WriteQueueImpl","level":"error","error":"`undefined` is not supported by the IPLD Data Model and cannot be encoded","msg":"Error processing task"}

I tried to be smart about it

## Summary by CodeRabbit

- **New Features**
	- Enhanced document update processing with integrated data sanitization. Document values are now validated and cleaned automatically, ensuring proper handling of special cases (null, undefined, non-numeric values, or malformed arrays) with errors returned for invalid inputs.

- **Tests**
	- Expanded test coverage confirms the updated behavior for edge-case values during document updates, including null, undefined, NaN, Infinity, and arrays containing invalid values.

### DIFF
--- a/src/crdt-helpers.ts
+++ b/src/crdt-helpers.ts
@@ -53,6 +53,35 @@ function toString<K extends IndexKeyType>(key: K, logger: Logger): string {
   }
 }
 
+export function sanitizeDocumentFields(obj: unknown): unknown {
+  if (Array.isArray(obj)) {
+    return obj.map((item: unknown) => {
+      if (typeof item === 'object' && item !== null) {
+        return sanitizeDocumentFields(item);
+      }
+      return item;
+    });
+  } else if (typeof obj === 'object' && obj !== null) {
+    const typedObj = obj as Record<string, unknown>;
+    const result: Record<string, unknown> = {};
+    for (const key in typedObj) {
+      if (Object.hasOwnProperty.call(typedObj, key)) {
+        const value = typedObj[key];
+        if (value === null || (!Number.isNaN(value) && value !== undefined)) {
+          if (typeof value === 'object' && key[0] !== '_') {
+            const sanitized = sanitizeDocumentFields(value);
+            result[key] = sanitized;
+          } else {
+            result[key] = value;
+          }
+        }
+      }
+    }
+    return result;
+  }
+  return obj;
+}
+
 export async function applyBulkUpdateToCrdt<T extends DocTypes>(
   store: StoreRuntime,
   tblocks: CarTransaction,
--- a/src/crdt-helpers.ts
+++ b/src/crdt-helpers.ts
@@ -53,22 +53,22 @@ function toString<K extends IndexKeyType>(key: K, logger: Logger): string {
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/crdt-helpers.ts
```diff
@@ -53,6 +53,35 @@ function toString<K extends IndexKeyType>(key: K, logger: Logger): string {
   }
 }
 
+export function sanitizeDocumentFields(obj: unknown): unknown {
+  if (Array.isArray(obj)) {
+    return obj.map((item: unknown) => {
+      if (typeof item === 'object' && item !== null) {
+        return sanitizeDocumentFields(item);
```
mabels, line 60: why call with undefined

src/crdt-helpers.ts
```diff
@@ -53,6 +53,35 @@ function toString<K extends IndexKeyType>(key: K, logger: Logger): string {
...
+      }
+      return item;
+    });
+  } else if (typeof obj === 'object' && obj !== null) {
+    const typedObj = obj as Record<string, unknown>;
```
mabels, line 65: this is also entered if it is a class instances

src/crdt-helpers.ts
```diff
@@ -53,6 +53,35 @@ function toString<K extends IndexKeyType>(key: K, logger: Logger): string {
...
+    const result: Record<string, unknown> = {};
+    for (const key in typedObj) {
+      if (Object.hasOwnProperty.call(typedObj, key)) {
+        const value = typedObj[key];
+        if (value === null || (!Number.isNaN(value) && value !== undefined)) {
```
mabels, line 70: why filter something with NaN?

src/crdt-helpers.ts
```diff
@@ -53,6 +53,35 @@ function toString<K extends IndexKeyType>(key: K, logger: Logger): string {
...
+          if (typeof value === 'object' && key[0] !== '_') {
```
mabels: startsWith

src/crdt.ts
```diff
@@ -104,6 +105,10 @@ export class CRDT<T extends DocTypes> {
   async bulk(updates: DocUpdate<T>[]): Promise<CRDTMeta> {
     await this.ready();
     const prevHead = [...this.clock.head];
+    updates = updates.map((dupdate: DocUpdate<T>) => ({
+      ...dupdate,
+      value: sanitizeDocumentFields(dupdate.value) as typeof dupdate.value
```
mabels: as T or better sanitize gets generic

src/crdt-helpers.ts
jchris: we only want to clean field names not array slots

src/crdt-helpers.ts
jchris: I think that should be OK, we just want to clean out undefined fields

src/crdt-helpers.ts
jchris: you get `NaN` from `parseInt("foo")`

PR #707, opened 2025-04-13
Improve index robustness

jchris:
This PR improves index robustness in two ways:

1. **Accept different map functions**: Instead of throwing errors when map functions don't match, we now accept the new function, reset the index, and continue operation with a warning. This helps with bundler issues where function string representations might change even when functionality is identical.

2. **Simplify document inclusion**: Removed the  property and simplified the logic to always include docs by default in query results.

All tests are passing across all environments (memory, file, and indexeddb).

## Summary by CodeRabbit

- **Refactor**
  - Streamlined query option handling to ensure documents are included consistently.
  - Enhanced data integrity by automatically resetting indexes when mapping functions change.
  - Improved error handling and readability in document fetching logic.
  
- **New Features**
  - Introduced a specific error handling mechanism for missing keys in CRDT functions.
  
- **Tests**
  - Added a new test suite to validate the behavior of the `useDocument` hook with non-existent document IDs.

### DIFF
--- a/src/crdt-helpers.ts
+++ b/src/crdt-helpers.ts
@@ -33,6 +33,7 @@ import {
 } from "./types.js";
 import { Logger } from "@adviser/cement";
 import { CarTransactionImpl } from "./blockstore/transaction.js";
+import { NotFoundError } from "./utils.js";
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 function time(tag: string) {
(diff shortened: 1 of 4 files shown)


jchris:
@mabels this cleans up cosmetic stuff for apps in the browser. I'm not sure about this use of the not found error. Maybe we should merge this and do another PR that applies similar error handling across the board?

PR #840, opened 2025-05-29
Jchris/public-true

jchris:
fixes the public: true implementation

## Summary by CodeRabbit

- **New Features**
	- Added support for configuring databases as public (unencrypted), ensuring all related store URIs use an "insecure" key and cryptography is disabled.
- **Bug Fixes**
	- Resolved issues with URI parameter handling and ensured correct defaulting of store keys and index names.
- **Tests**
	- Introduced comprehensive tests for public database configuration and URI default logic, including no-crypto behavior verification.
- **Refactor**
	- Simplified and clarified URI defaulting logic, especially for public/private database options.
- **Documentation**
	- Added a planning and progress note outlining the refactor and next steps for public database support.

### DIFF
--- a/src/runtime/keyed-crypto.ts
+++ b/src/runtime/keyed-crypto.ts
@@ -195,7 +195,7 @@ class noCrypto implements CryptoAction {
       id: sthis.nextId().str,
       name: "noCrypto",
       get: () => {
-        throw this.logger.Error().Msg("noCrypto.get not implemented").AsError();
+        return Promise.resolve(undefined); // No specific key for noCrypto
       },
       upsert: () => {
         throw this.logger.Error().Msg("noCrypto.upsert not implemented").AsError();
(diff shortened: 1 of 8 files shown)


src/runtime/keyed-crypto.ts
```diff
@@ -195,7 +195,7 @@ class noCrypto implements CryptoAction {
       id: sthis.nextId().str,
       name: "noCrypto",
       get: () => {
-        throw this.logger.Error().Msg("noCrypto.get not implemented").AsError();
+        return Promise.resolve(undefined); // No specific key for noCrypto
```
jchris, line 198: not sure this is right... I think we want to point to the 'insecure' key here? or maybe this code path should not exercise when key == 'insecure' in which case we can revert this line. investigating.

jchris: I checked, this change is required

mabels:
I will close this --- it's a completely misguided approach to solve a problem.

We should define the requirements more closely and then approach the problem from scratch.

PR #842, opened 2025-05-30
Jchris/dashboard-api

jchris:


jchris:
If this markdown file is correct, it's probably a useful reference. @mabels if you just put code review comments in, I'll do the cleanup

mabels:
We should at least not publish "fake news" how urgent is that?

I need at least 2-5 days to produce qualitative feedback.

PR #844, opened 2025-05-31
Jchris/error-messages

jchris:


mabels:
Pls do not introduce try/catch block in exception2Result blocks.
And think about the package size --- if you want extensive error messages we should build something with error codes and have in the long text somewhere.

PR #889, opened 2025-06-21
feat: add SQLite database support and dashboard development configuration

jchris:


mabels:
What is this?
I don't get it.

PR #1246, opened 2025-10-10
feat: add shareWithUser endpoint

necrodome:
## Summary by CodeRabbit

* **New Features**
  * Users can share ledgers with others via email.
  * Configurable roles and permissions when granting access.
  * Prevents sharing a ledger with oneself.

* **Security**
  * Improved token verification and session/token refresh handling for share actions.
  * Clearer authorization and not-found error messages during sharing flows.

### DIFF
--- a/dashboard/backend/api.ts
+++ b/dashboard/backend/api.ts
@@ -1846,7 +1846,19 @@ export class FPApiSQL implements FPApiInterface {
       return Result.Err("Cannot share ledger with yourself");
     }
 
-    // 6. Add user to ledger (no need to add to tenant for sharing)
+    // 6. Add user to tenant first
+    const rAddUserToTenant = await this.addUserToTenant(this.db, {
+      userName: req.email,
+      tenantId: ledger.tenantId,
+      userId: targetUser.userId,
+      role: "member",
+    });
+
+    if (rAddUserToTenant.isErr()) {
+      return Result.Err(rAddUserToTenant.Err());
+    }
+
+    // 7. Add user to ledger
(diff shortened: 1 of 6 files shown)


dashboard/backend/api.ts
```diff
@@ -1709,6 +1712,109 @@ export class FPApiSQL implements FPApiInterface {
...
+    const targetUser = users[0];
+
+    // Prevent self-sharing
+    if (targetUser.userId === userId) {
+      return Result.Err("Cannot share ledger with yourself");
+    }
+
+    // 6. Add user to tenant first
+    const rAddUserToTenant = await this.addUserToTenant(this.db, {
+      userName: req.email,
```
mabels, line 1851: you want only share a ledger not a hole tenant --- that basically shares all ledger of the user to the other.

mabels (changes requested):
Where you do an invite? --- it does not help

mabels (changes requested):
From where the double CORS is coming from?
 --- this is not correct, this will break the local version.

necrodome:
> From where the double CORS is coming from? --- this is not correct, this will break the local version.

https://github.com/fireproof-storage/fireproof/blob/b957cc6b95feb0e547db1f83456b78cf07d3889b/dashboard/backend/cf-serve.ts#L38

mabels:
that had been there for ever, the question where is the second coming from

mabels:
pls use u github handle in the branch name @necrodome

mabels:
move the CORS fix in mabels/fix-double-cors or #1257

PR #1280, opened 2025-10-23
feat: add loaded flag to React hooks

jchris:
## Summary

Add a `loaded` boolean property to all React hook results to distinguish between:
- **Not loaded yet**: `loaded=false`, data is in initial/empty state
- **Loaded**: `loaded=true`, data reflects actual query results (may be empty)

This solves a fundamental UX problem where hooks initially return empty data structures before the first fetch completes, making it impossible to distinguish between "loading" and "legitimately empty results".

### Core Implementation
- Created base `HookResult` interface with `readonly loaded: boolean`
- Updated all hook result types to extend `HookResult`:
  - `LiveQueryResult`
  - `AllDocsResult`
  - `ChangesResult`
  - `UseDocumentResultObject`

### Hook Implementations
- **useLiveQuery**: Tracks loaded state, sets to `true` after first `refreshRows()` completes
- **useDocument**: Tracks loaded state, sets to `true` after first `refresh()` completes
- **useAllDocs**: Tracks loaded state, sets to `true` after first `refreshRows()` completes
- **useChanges**: Tracks loaded state, sets to `true` after first `refreshRows()` completes

### Testing
- Added comprehensive test suite for `loaded` behavior
- Tests validate loaded flag correctly distinguishes empty vs not-yet-loaded states
- All existing tests continue to pass

## Usage

```tsx
const { rows, docs, loaded } = useLiveQuery("type", { key: "todo" });
const { doc, loaded, save } = useDocument({ title: "" });
const { docs, loaded } = useAllDocs();
const { docs, loaded } = useChanges([], {});

if (!loaded) {
  return <LoadingSpinner />;
}

// Now we know we have real data (or confirmed empty results)
return <TodoList items={docs} />;
```

## Test Plan

- [x] All existing tests pass
- [x] New tests verify loaded flag starts false
- [x] New tests verify loaded flag becomes true after data loads
- [x] Test validates empty results with loaded=true are distinguishable from not-yet-loaded
- [x] Tests cover all four hooks: useLiveQuery, useDocument, useAllDocs, useChanges
- [x] TypeScript diagnostics are clean

## Breaking Changes

None - this is a purely additive change. The `loaded` property is added to existing return types.

## Summary by CodeRabbit

* **New Features**
  * Hooks for live queries, document, all-docs, and changes now expose a readonly hydrated flag; hydration resets when the target document, query, or parameters change.

* **Bug Fixes / API**
  * Attach behavior updated to return attach state via a callable instead of a static value.

* **Tests**
  * Added/expanded suites validating hydrated semantics and reset-on-change flows; test lifecycles consolidated and timeouts shortened.

* **Chores**
  * Reduced noisy logging and tightened some startup/connection closing behavior.

### DIFF
--- a/cloud/backend/base/msg-dispatcher-impl.ts
+++ b/cloud/backend/base/msg-dispatcher-impl.ts
@@ -135,7 +135,7 @@ export function buildMsgDispatcher(_sthis: SuperThis /*, gestalt: Gestalt, ende:
             if (qsidEqual(conn, msg.conn)) {
               if (msg.message.startsWith("/close-connection")) {
                 setTimeout(() => {
-                  item.ws?.close();
+                  item.ws?.close(4711, "closed by /close-connection");
                   ctx.wsRoom.removeConn(...item.conns);
                 }, 50);
               }
--- a/cloud/backend/cf-d1/setup-backend-d1.ts
+++ b/cloud/backend/cf-d1/setup-backend-d1.ts
@@ -31,7 +31,7 @@ export async function setupBackendD1(
     if (mightPid) {
       pid = +mightPid;
     }
-    if (chunk.includes("Starting local serv")) {
+    if (chunk.includes("Ready on")) {
       waitReady.resolve(true);
     }
   });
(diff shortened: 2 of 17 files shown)


jchris:
The lifecycle is per query / hook, not per db, so this interface makes sense.

mabels (changes requested):
i don't like the loaded attribute name --- it does not explain that this is react related.

PR #1467, opened 2025-12-15
feat: shared ledger access for invited users

jchris:
## Summary

When an invited user calls `ensureCloudToken` with an appId, they currently get a new ledger created for them (with name `{appId}-{userId}`), causing a UNIQUE constraint violation if they're invited to the owner's ledger.

This PR adds prefix matching: before creating a new ledger, check if the user has access to any ledger whose name **starts with** the appId. This allows invited users to access the owner's ledger instead of creating a duplicate.

## Question for review

**Should we use prefix matching, or should we remove the user-id suffix from ledger names entirely?**

### Option A: Prefix matching (this PR)
- Ledger name format stays: `{appId}-{ownerUserId}`  
- On `ensureCloudToken`, check for ledgers matching `LIKE '{appId}%'`
- Pros: preserves user-specific ledger names, works with existing data
- Cons: adds query complexity, relies on naming convention

### Option B: Remove user-id suffix
- Ledger name format becomes just: `{appId}`
- Simpler logic, but name collision risk if same appId used by different apps
- Would need migration for existing ledgers

## Test plan
- [x] Added test: "ensureCloudToken grants invited user access to shared ledger via prefix match"
- [x] All 34 existing tests pass

## Summary by CodeRabbit

* **New Features**
  * Reuse existing app-scoped ledger bindings when available; invited users can be auto-redeemed and gain access without creating a new ledger. New ledgers are named after the app ID and binding limits are enforced.

* **Bug Fixes**
  * Avoid duplicate membership insertion by treating existing ledger memberships as successful.

* **Refactor**
  * Streamlined user lookup/condition building and tightened authorization checks.

* **Tests**
  * Expanded coverage for invite redemption, shared-ledger access, prefix-matching, and unauthorized-access scenarios.

### DIFF
--- a/dashboard/backend/api.ts
+++ b/dashboard/backend/api.ts
@@ -155,12 +155,15 @@ export class FPApiSQL implements FPApiInterface {
     return this.#checkAuth(req, (req) => getCertFromCsr(this.ctx, req));
   }
   async ensureCloudToken(req: ReqEnsureCloudToken, ictx: Partial<FPTokenContext> = {}): Promise<Result<ResEnsureCloudToken>> {
-    // For new users, ensureUser must be called first to create them
+    // p0.46 - For new users, ensureUser must be called first to create them
     // This allows invited users to access shared ledgers on first visit
+    console.log("[ensureCloudToken p0.46] starting for appId:", req.appId);
     const rUser = await ensureUser(this.ctx, { type: "reqEnsureUser", auth: req.auth });
     if (rUser.isErr()) {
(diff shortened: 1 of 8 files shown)


jchris:
Not sure if this is the best way. The underlying issue is that a user responding to an invite is getting bound to a new ledger not the ledger from the source user. This fixes that -- not sure its the best way to fix it

jchris:
I deployed this but I dont think its what we want

mabels:
it's not needed any more --- thats all move into vibes.diy.

PR #1496, opened 2025-12-30
fix(security): remove storekey=insecure encryption bypass

bbopen:
## Summary
- Removes the `storekey=insecure` option that bypassed encryption
- All data is now encrypted by default
- Tests updated to verify rejection or use proper encryption

## Changes
- `ensureKeyFromUrl()` returns error for `storekey=insecure`
- `keyedCryptoFactory()` throws for `storekey=insecure`
- Removed unused `noCrypto` and `nullCodec` classes
- Updated tests to use behavior-based assertions instead of exact CID checks

## Test plan
- [x] Tests verify insecure mode is rejected with helpful error
- [x] Existing tests updated to use encryption
- [x] `pnpm format` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (128 files, 1775 tests)
- [x] `pnpm build` passes

## Breaking Change
The `storekey=insecure` URL parameter is no longer supported. Remove it from any URLs to use automatic key generation.

Fixes ROBUST-03

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced encryption requirement by rejecting unencrypted storage configuration. The system now rejects attempts to use unencrypted data storage and requires all data to be encrypted, returning a clear error message when unencrypted storage is requested.

### DIFF
--- a/core/keybag/key-bag.ts
+++ b/core/keybag/key-bag.ts
@@ -85,11 +85,29 @@ export class KeyBag implements KeyBagIf {
     );
   }
 
+  /**
+   * Ensures a store key is present in the URL, generating one if needed.
+   * Rejects attempts to use insecure (unencrypted) storage.
+   *
+   * @param url - The URI to ensure has a store key
+   * @param keyFactory - Factory function to generate a key name if none exists
+   * @returns Result containing the URI with store key, or error if insecure mode attempted
+   * @throws Error if storekey=insecure is attempted (removed for security)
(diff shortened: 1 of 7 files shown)


mabels:
hi, we want to keep this for now --- for debugging. It has to be enabled explicitly.
In the next architecture, the encryption is moved from the current core into the gateway implementation, and then the responsibility is pushed to the gateway implementation.

PR #1583, opened 2026-02-01
fix: CID serialization in blob URL construction

popmechanic:
## Summary
Fixes CID objects being serialized as `[object Object]` in blob URLs, causing 404 errors on GET requests.

## Root Cause
When CIDs are deserialized from cloud metadata via dag-json, they arrive as plain objects `{"/": "bafy..."}` instead of proper CID instances. Calling `.toString()` on these returns `[object Object]`.

## Fix
Handle multiple CID formats when constructing URLs:
- String keys → pass through
- CID objects with `.toString()` → call toString() if it doesn't return `[object Object]`
- dag-json link objects `{"/": "bafy..."}` → extract "/" property
- Fallback to String()

## Files Changed
- `core/blockstore/store.ts` - Client-side CID stringification in DataStoreImpl.load()
- `cloud/backend/base/pre-signed-url.ts` - S3 pre-signed URL construction

## Testing
Verified on https://quicknotes2.exe.xyz:
- Notes persist across page refreshes ✅
- Blob GET requests return 200 with proper CID URLs ✅
- No `[object Object]` in network requests ✅


popmechanic:
Closing - wrong approach. Branch divergence from main causes massive diff.

PR #1593, opened 2026-02-07
fix(docker): nginx CORS, WebSocket /fp, upload limit, BLOB_PROXY_URL

popmechanic:
## Summary

Four fixes to the Docker nginx proxy and startup script that affect production deployments:

- **Strip duplicate CORS headers** — upstream backends (dashboard, cloud-backend) add their own CORS headers. When an outer reverse proxy (Caddy, Cloudflare, custom nginx) also adds CORS, browsers reject the response due to duplicate `Access-Control-Allow-Origin`. `proxy_hide_header` strips them at the nginx layer so deployers can control CORS at one place.

- **Add WebSocket upgrade to `/fp`** — Fireproof clients connect to `/fp` for real-time sync via WebSocket, but the nginx config only had `Upgrade`/`Connection` headers on `/ws`. Without these headers on `/fp`, WebSocket connections silently fall back to polling or fail entirely.

- **Raise `client_max_body_size` to 100m** — nginx defaults to 1MB, which causes `413 Payload Too Large` errors when uploading blobs or images over 1MB.

- **Add `BLOB_PROXY_URL` to generated `.env`** — `start.sh --setup` generates `.env` but omits `BLOB_PROXY_URL`. On remote/VM deployments, the cloud-backend generates blob download URLs using its internal default (`http://localhost:8080`), which clients can't reach. Adding this variable lets deployers override it with their public URL.

## Test plan

- [ ] `./docker/start.sh --setup` generates `.env` containing `BLOB_PROXY_URL`
- [ ] `BLOB_PROXY_URL=https://example.com ./docker/start.sh --setup` propagates the override
- [ ] `./docker/start.sh --verify` passes all routes including `/fp` WebSocket
- [ ] Upload a file >1MB through the proxy without 413 error
- [ ] Confirm no duplicate CORS headers when behind an outer proxy

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -23,6 +23,10 @@ http {
         listen 8080;
         server_name _;
 
+        # Allow large file uploads (blobs, images).
+        # nginx defaults to 1MB which causes 413 errors.
+        client_max_body_size 100m;
+
         # Strip CORS headers from upstream backends to prevent duplication.
         # The dashboard and cloud-backend services add their own CORS headers,
         # which causes "multiple values" errors when an outer reverse proxy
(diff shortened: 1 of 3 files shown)


popmechanic:
This fix has been tested on the exe.dev environment. It will probably need additional testing against the new CloudFlare deployment.

popmechanic:
[Write-up on the sync refresh workaround](https://gist.github.com/popmechanic/c6b7ab4ab80465e3b08f0685c63f815b). 

@mabels, curious to hear your thoughts on Opus 4.6's diagnosis of the upstream fireproof issue:

**Why This is a Workaround**
The proper fix belongs in core/blockstore/loader.ts. The handleMetaStream() function processes incoming sync metadata but never calls ledger._notify() or fires clock events that would trigger subscriptions. The existing test file core/tests/fireproof/attachable-subscription.test.ts documents this exact gap — subscriptions fire for local writes but not for remotely synced data.

The poll workaround is appropriate for the clerk hook because:

It's cheap (a few IndexedDB reads)
It's self-limiting (stops once sync settles)
It doesn't touch internals or private APIs
It works regardless of how the core eventually fixes the notification gap
When the core properly notifies subscribers on sync, this poll becomes a no-op — allDocs() still runs a few times but the data will already be visible via subscriptions, and the stable count exits quickly.

PR #1720, opened 2026-03-28
fix(core): batch write queue to fix useLiveQuery concurrent write reactivity

jchris:
## Summary

- **Merged write queue chunk processing** into a single `crdt.bulk()` call instead of parallel per-item worker calls. This ensures one transaction and one subscription notification per chunk, fixing the race where early notifications fired before sibling writes completed.
- **Gated ledger notifications** on `writeQueue.hasPending()` to suppress intermediate chunk notifications — only the final chunk fires.
- **Added property-based test** (fast-check, N=1..50) that finds the minimal failing batch size, plus a regression test for the original 12-doc case.

## Test plan

- [x] Property test passes: `useLiveQuery` reflects all N docs for N=1..50
- [x] Regression test passes: 12 concurrent puts all visible
- [x] All 1607 core tests pass (0 failures)
- [x] All 55 use-fireproof hook tests pass (0 failures)
- [x] `pnpm check` passes

## Summary by CodeRabbit

* **New Features**
  * Added ability to detect pending write operations.

* **Bug Fixes**
  * Prevented no-update notifications while writes are still pending.
  * Improved live-query consistency and ordering during concurrent writes.

* **Tests**
  * Added concurrent-write tests for live queries and payload subscriptions (including property-based tests).
  * Added test dependency to support property-based testing.

### DIFF
--- a/core/base/ledger.ts
+++ b/core/base/ledger.ts
@@ -269,9 +269,6 @@ class LedgerImpl implements Ledger {
 
   private async _notify(updates: DocUpdate<DocTypes>[]) {
     await this.ready();
-    if (this.writeQueue.hasPending()) {
-      return;
-    }
     if (this._listeners.size) {
       const docs: DocWithId<DocTypes>[] = updates.map(({ id, value }) => ({ ...value, _id: id }));
       for (const listener of this._listeners) {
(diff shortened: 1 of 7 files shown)


mabels:
What is the motivation about this?

It's fairly complex --- i need about 1-2days to weigh that.

jchris:
regarding the test coverage stuff, I am trying to track down stuff I experience when using apps and capture it as test cases. we can hold on reviewing it for a while.

